PR #3, opened 2024-01-26
fix: TM-538 add new user to organisation after accept invite to monitoring project

pachonjcl:
There was a missing piece of logic that now we need to add a new user into the organization of the project that invited to monitor.

Commits: 
- fix: TM-538 add missing logic for users outside TM signing up
- fix: TM-538 deleted imports

### DIFF
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -94,32 +94,13 @@ class AuthController extends Controller
         $verification = VerificationModel::where('token', '=', $data['token'])
             ->where('user_id', '=', $me->id)
             ->firstOrFail();
-            $verification->delete();
+
         $me->email_address_verified_at = new DateTime('now', new DateTimeZone('UTC'));
+        $me->saveOrFail();
+        $verification->delete();
+
         return JsonResponseHelper::success((object) [], 200);
     }
-
-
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/AuthController.php
```diff
@@ -130,12 +107,20 @@ public function verifyUnauthorizedAction(VerifyRequest $request): JsonResponse
...
-        $invites = ProjectInvite::where('email_address', $user->email_address)->get();
+        $invites = ProjectInvite::where('email_address', $user->email_address)->orderBy('created_at', 'desc')->get();
+        $is_first = true;
         foreach ($invites as $invite) {
             $invite = ProjectInvite::where('token', $invite->token)
                 ->where('email_address', $user->email_address)
                 ->first();
+            if ($is_first) {
+                $organisation_id = $invite->project->organisation_id;
+                $user->organisation_id = $organisation_id;
```
roguenet, line 120: Not a big deal, but it's a little weird to me to stash the organisation ID in an intermediate variable that's not used elsewhere.

roguenet (approved):

app/Http/Controllers/AuthController.php
pachonjcl: done

PR #6, opened 2024-01-26
TM-534_custom_forms_not_translated

pachonjcl:
Use `lang` query param to set the locale on the response for custom forms entity endpoint


app/Http/Controllers/V2/Nurseries/ViewNurseryWithFormController.php
```diff
@@ -7,13 +7,18 @@
...
+use Illuminate\Support\Facades\App;
 
 class ViewNurseryWithFormController extends Controller
 {
     public function __invoke(Request $request, Nursery $nursery): NurseyWithSchemaResource
     {
         $this->authorize('read', $nursery);
 
+        if ($request->query('lang')) {
+            App::setLocale($request->query('lang'));
```
roguenet, line 19: This seems like a pattern we want more generally applied. I don't know what the idiomatic pattern would be in Laravel... maybe a concern that gets mixed in at the base `Controller` level?

pachonjcl: I created a middleware that would handle this repetitive task.
you would just need to apply `->middleware('i18n')` in the end of the route.
This common pattern was repeated in many places.

roguenet (commented):
That's a nice cleanup, thanks. One more thought: Is there any downside to including this middleware in all routes? I can't imagine it would incur much/any performance penalty on a route that doesn't happen to have any translatable text, and this way we just don't have to worry about it on the BE anymore.

pachonjcl:
My only concern right now is that we are using mostly GET endpoints with this middleware, because we obtain the translated labels without the risk of updating this by mistake.
But if we send a update request with the locale, it is not clear if we can alter the wrong value for forms titles, questions and helpers.

In the future if we are sure of that, we can add the middleware on all routes.

PR #8, opened 2024-01-29
chore: php code style fix

pachonjcl:
We didn't use the lint that is on the make file.

This commit should fix all current code style missing.


roguenet (approved):
👍🏻 I'll add a note to have github do linting on PRs so we don't miss changes like this in PRs in the future.

PR #24, opened 2024-02-08
feat: TM-557 conditional fields saving

LimberHope:
[task link](https://gfw.atlassian.net/browse/TM-557)

### DIFF
--- a/app/Models/Traits/UsesLinkedFields.php
+++ b/app/Models/Traits/UsesLinkedFields.php
@@ -11,24 +11,16 @@ trait UsesLinkedFields
         $localAnswers = [];
         foreach ($this->form->sections as $section) {
             foreach ($section->questions as $question) {
-                if ($question->input_type === 'conditional') {
-                    $localAnswers[$question->uuid] = data_get($input, $question->uuid);
-                } else {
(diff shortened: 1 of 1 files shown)


app/Models/Traits/UsesLinkedFields.php
```diff
@@ -11,18 +11,22 @@ public function updateAllAnswers(array $input): array
...
+                            'organisation' => $this->organisation,
+                            'project-pitch' => $this->projectPitch,
+                        ]);
 
-                    if (! empty($linkedFieldInfo)) {
-                        $this->updateLinkedFieldValue($linkedFieldInfo, data_get($input, $question->uuid));
+                        if (! empty($linkedFieldInfo)) {
+                            $this->updateLinkedFieldValue($linkedFieldInfo, data_get($input, $question->uuid));
+                        }
                     }
```
pachonjcl, line 30: I think the code handles too many conditional and we can apply "early return" or "early exit" pattern.

Meaning that instead of handling cases in this way:
```
if (condition) {
//something
} else {
//something else 
}
```

We can apply something like 
```
if (!condition) {
//something else
   continue;
 }
 //something
 ```

cesarLima1 (approved):

PR #28, opened 2024-02-14
[TM-559] Task/Report database associations

roguenet:
https://gfw.atlassian.net/browse/TM-559

In order to improve the performance and data consistency of reports and tasks, a has one / belongs to relationship has been established between Task and ProjectReport, and a has many / belongs to relationship has been established between Task and both SiteReport and NurseryReport.


cesarLima1 (approved):
Everything looks good.
Just to leave a reminder, we would have to create a new ticket to correct the dueDate in the new CreateTaskDueJob.php considering that nothing was modified and that when they ran it in December, the records were generated but not with the correct dueDate.

roguenet:
> Everything looks good. Just to leave a reminder, we would have to create a new ticket to correct the dueDate in the new CreateTaskDueJob.php considering that nothing was modified and that when they ran it in December, the records were generated but not with the correct dueDate.

Can you elaborate on this? From the (code)[https://github.com/wri/wri-terramatch-api/pull/28/files#diff-003ec0c6add7a92febf3f63553c73623180cb23adf8b5b01f0458ab06e766f6dL56], I thought that the Tasks were being created with the same due date as the project report in the previous generation of the generation tools.

roguenet:
> > Everything looks good. Just to leave a reminder, we would have to create a new ticket to correct the dueDate in the new CreateTaskDueJob.php considering that nothing was modified and that when they ran it in December, the records were generated but not with the correct dueDate.
> 
> Can you elaborate on this? From the [code](https://github.com/wri/wri-terramatch-api/pull/28/files#diff-003ec0c6add7a92febf3f63553c73623180cb23adf8b5b01f0458ab06e766f6dL56), I thought that the Tasks were being created with the same due date as the project report in the previous generation of the generation tools.

@cesarLima1 In fact, the whole system as I understand it wouldn't work if the Tasks were generated with the wrong due date, because that's how they're all linked together today. Maybe I'm misunderstanding what you're referring to?

cesarLima1:
What I'm pointing out is these lines that weren't updated in CreateTaskDueJob.php:
`    public function __construct(string $frameworkKey, int $dueMonth = null)
    {
        $this->frameworkKey = $frameworkKey;
        if ($dueMonth) {
            $carbonDate = Carbon::createFromFormat('m', $dueMonth);
            $this->dueDate = $carbonDate->isPast() ? $carbonDate->addYear()->firstOfMonth(5) : $carbonDate->firstOfMonth(5);
        } else {
            $this->dueDate = Carbon::now()->addMonth()->startOfDay();
        }
    }`
As I mentioned, when it was run in December, it did generate a due date, but it wasn't created with the correct day or time. That's why I suggest creating a dedicated ticket for this task.

PR #36, opened 2024-02-19
feat: TM-567 has toggle saving in custom form

LimberHope:
[task link](https://gfw.atlassian.net/browse/TM-567)

### DIFF
--- a/app/Http/Controllers/V2/Forms/UpdateFormController.php
+++ b/app/Http/Controllers/V2/Forms/UpdateFormController.php
@@ -38,6 +38,17 @@ class UpdateFormController extends Controller
             }
         }
 
+        foreach (data_get($updateFormRequest, 'form_sections', []) as $formSection) {
+            $currentFormSection = $this->updateFormSection($formSection, $form);
+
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/V2/Forms/UpdateFormController.php
```diff
@@ -38,6 +38,17 @@ public function __invoke(Form $form, UpdateFormRequest $updateFormRequest): Form
...
         }
 
+        foreach (data_get($updateFormRequest, 'form_sections', []) as $formSection) {
+            $currentFormSection = $this->updateFormSection($formSection, $form);
+
+            foreach (data_get($formSection, 'form_questions', []) as $formQuestion) {
+                $this->handleQuestions($formQuestion, $currentFormSection);
+                if (data_get($formQuestion, "input_type") == "select") {
+                    $this->updateFormQuestion($formQuestion, $currentFormSection);
+                }
```
pachonjcl, line 48: There is a double for-each loop above doing the same thing, except for this conditional.

Wouldn't it be better to handle this on the above loop?

pachonjcl (changes requested):

PR #38, opened 2024-02-22
fix: TM-575 add new columns

cesarLima1:
add new columns based on ticket [TM-575](https://gfw.atlassian.net/browse/TM-575)


pachonjcl (approved):
LGTM

PR #41, opened 2024-02-23
Tm 662 implement total sections for terrafund dashboard header

LimberHope:
[task link](https://gfw.atlassian.net/browse/TM-662)

### DIFF
--- a/app/Http/Controllers/V2/Dashboard/TotalTerrafundHeaderDashboardController.php
+++ b/app/Http/Controllers/V2/Dashboard/TotalTerrafundHeaderDashboardController.php
@@ -2,6 +2,7 @@
 
 namespace App\Http\Controllers\V2\Dashboard;
 
+use App\Helpers\TerrafundDashboardQueryHelper;
 use App\Http\Controllers\Controller;
 use App\Models\V2\Projects\Project;
 use App\Models\V2\Projects\ProjectReport;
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/V2/Dashboard/TotalTerrafundHeaderDashboardController.php
```diff
@@ -0,0 +1,143 @@
...
+    public function getTotalTreesGrownGoalSum(Request $request)
+    {
+        $projects = Project::where('framework_key', 'terrafund');
+        if ($request->has('country')) {
+            $country = $request->input('country');
+            $projects->where('country', $country);
+        } elseif ($request->has('uuid')) {
+            $projectId = $request->input('uuid');
+            $projects->where('uuid', $projectId);
+        }
```
pachonjcl, line 138: This is a repeated in some of the methods, please can you add a helper function to simplify this?

For example a helper that would read all query params and then return the query builder.

pachonjcl (changes requested):

app/Http/Controllers/V2/Dashboard/TotalTerrafundHeaderDashboardController.php
```diff
@@ -0,0 +1,120 @@
...
+            'total_non_profit_count' => $this->getTotalNonProfitCount($request),
+            'total_enterprise_count' => $this->getTotalEnterpriseCount($request),
+            'total_entries' => $this->getTotalJobsCreatedSum($request),
+            'total_hectares_retored' => $this->getTotalHectaresRestoredGoalSum($request),
+            'total_trees_restored' => $this->getTotalTreesRestoredSum($request),
+            'total_trees_restored_goal' => $this->getTotalTreesGrownGoalSum($request),
+        ]);
+    }
+
+    public function buildQueryFromRequest($query, $request)
```
pachonjcl, line 27: still missing the `where('framework_key', 'terrafund')` that is repeated in all calls right?

pachonjcl (changes requested):

app/Http/Controllers/V2/Dashboard/TotalTerrafundHeaderDashboardController.php
pachonjcl: and the method should go into another file as a helper, so we can import it in other endpoints.

pachonjcl (approved):

PR #43, opened 2024-02-23
[TM-561] Unify status logic

roguenet:
https://gfw.atlassian.net/browse/TM-561
FE PR: https://github.com/wri/wri-terramatch-website/pull/71

This was one of the big cleanup passes in this epic, and involved a lot of learning Laravel to get the `ReportModel` interface to interact correctly with controller parameter binding.  There are three main improvements:
 * The `status` field on reports is now driven by a state machine, which adds some guardrails around what sorts of changes are allowed
 * All interactions with the `status` field are driven by the same controller regardless of the type of report being updated, so there is a lot less copy pasted code floating around that would need updating when we have changes to make.
 * The `completion_status` DB column was totally superfluous. It wasn't really being used on the FE (although I plan to fix that, so I kept it in the API response), and there was no reason it couldn't be calculated at API response render time from the `completion` field.


openapi-src/V2/paths/Forms/post-v2-forms-entity-form-uuid.yml
```diff
@@ -13,7 +13,7 @@ parameters:
     name: ENTITY
     in: path
     required: true
-    description: allowed values projects/sites/nurseries/project-reports/site-reports/nursery-reports
+    description: allowed values projects/sites/nurseries
```
roguenet, line 16: Creating reports via form isn't supported, and wasn't being used by the FE. This is simply a documentation update to reflect that fact.

PR #55, opened 2024-02-29
feat: TM-669 add endpoint for dashboard openapi

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-669)


LimberHope (approved):

cesarLima1 (approved):

app/Http/Controllers/DashboardDocumentationController.php
```diff
@@ -0,0 +1,31 @@
...
+namespace App\Http\Controllers;
+
+use Illuminate\Http\Request;
+use Illuminate\Http\Response;
+
+class DashboardDocumentationController extends Controller
+{
+    public function readAsYamlAction(Request $request): Response
+    {
+        $this->authorize('yes', 'App\\Models\\Default');
```
roguenet, line 12: What does this check accomplish?

resources/docs/dashboard.yml
```diff
@@ -0,0 +1,59 @@
+openapi: 3.0.0
+info:
+  title: Dashboard Terrafund API
```
roguenet, line 3: Since this is part of the existing `v2` API namespace, it seems to make sense to include this documentation in the big swagger doc that already exists. This feels more like an extension of the existing API, rather than a whole new API that needs separate documentation.

pachonjcl: The intention with having another OpenAPI endpoint only for dashboard is to make separate code generations bases. Otherwise website repository will be generating code for dashboard endpoints which doesn't use and geo repository will be generating code for all the other endpoints which are not going to be use.

app/Http/Controllers/DashboardDocumentationController.php
pachonjcl: It was a temporal placeholder for permissions to set later, but I will remove for now.

roguenet (approved):

roguenet (approved):

PR #56, opened 2024-02-29
feat: TM added endpoint for active projects and project list view

LimberHope:
Task Link in documentation
[Task Link](https://gfw.atlassian.net/wiki/spaces/TerraMatch/pages/976060417/5.+Project+List+View)
[Task Link](https://gfw.atlassian.net/wiki/spaces/TerraMatch/pages/937492508/3.+Dashboard+Country+Level)


app/Http/Controllers/V2/Dashboard/ActiveProjectsTableController.php
```diff
@@ -0,0 +1,150 @@
...
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\DB;
+
+class ActiveProjectsTableController extends Controller
+{
+    public function __invoke(Request $request)
+    {
+        return response()->json([
+            'project_list' => $this->getAllProjects($request),
+        ]);
```
roguenet, line 22: I believe the normal pattern here would be to have a resource class like the ProjectReportResource, and let it handle including the correct items for the API response.

app/Http/Controllers/V2/Dashboard/ActiveProjectsTableController.php
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    public function getAllProjects($request)
+    {
+        $projects = Project::query();
+        $projects = TerrafundDashboardQueryHelper::buildQueryFromRequest($projects, $request)
+            ->pluck('id')->toArray();
+        $activeProjects = [];
+        foreach ($projects as $projectId) {
```
roguenet, line 31: This controller is issuing a _ton_ of superfluous DB queries. You're pulling the ids for a bunch of projects and then iterating on those ids and in a lot of cases pulling the whole record, or part of the record from the DB in the code below. This is the definition of N+1 queries, and this kind of pattern is one of the main reasons the terramatch BE is so slow today. 

Two ideas:
* Just pull all the records initially instead of just the IDs. Then at least you're skipping most of the additional queries below.
* If you feel this really is the best pattern, then the additional queries should be done in bulk, which is a much more substantial refactor of this method. I honestly wouldn't do it this way because it's complicated and error-prone, but you could, for instance, pull the list of all names in one query with `Project::whereIn('id', $projects)->pluck('name')` instead of going back to the database for every single id in the initial pull.

roguenet (changes requested):

app/Helpers/TerrafundDashboardQueryHelper.php
```diff
@@ -2,11 +2,17 @@
...
+use App\Models\V2\Projects\Project;
+
 class TerrafundDashboardQueryHelper
 {
-    public static function buildQueryFromRequest($query, $request)
+    public static function buildQueryFromRequest($request)
     {
+        $query = Project::query();
         $query = $query->where('framework_key', 'terrafund');
+        $query = $query->whereHas('organisation', function ($query) {
```
roguenet, line 13: I left this feedback on the other PR as well, and it's a minor point, but this would be a little more idiomatic not saving off the query value on each step
```
$query = Project::where('framework_key', 'terrafund')
  ->whereHas('organisation', function ($query) {
```

app/Http/Controllers/V2/Dashboard/ActiveProjectsTableController.php
```diff
@@ -0,0 +1,80 @@
...
+                'date_added' => $project->created_at,
+            ];
+        }
+
+        return $activeProjects;
+    }
+
+    public function treesUnderRestoration($project)
+    {
+        return $project->sites()->get()->sum(function ($site) {
```
roguenet, line 47: You can avoid the `get()` in cases like this by accessing it as an attribute. Feel free to integrate this feedback on the other PR as well. It doesn't have a performance implication, but it's a little more readable:
```
return $project->sites->sum(function $site) {
```

roguenet (approved):
My only feedback this round is cosmetic. I do think a resource class would be better than returning raw json on the response, but I'm OK with letting this go as is, or can approve again after changes.

pachonjcl (approved):

PR #58, opened 2024-03-01
fix: endpoints logic improved

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-662)
[Task Link](https://gfw.atlassian.net/browse/TM-665)
[Task Link](https://gfw.atlassian.net/browse/TM-668)
[Task Link](https://gfw.atlassian.net/browse/TM-667)

### DIFF
--- a/app/Helpers/TerrafundDashboardQueryHelper.php
+++ b/app/Helpers/TerrafundDashboardQueryHelper.php
@@ -2,11 +2,17 @@
 
 namespace App\Helpers;
 
+use App\Models\V2\Projects\Project;
+
 class TerrafundDashboardQueryHelper
 {
-    public static function buildQueryFromRequest($query, $request)
+    public static function buildQueryFromRequest($request)
     {
+        $query = Project::query();
         $query = $query->where('framework_key', 'terrafund');
+        $query = $query->whereHas('organisation', function ($query) {
+            $query->whereIn('type', ['for-profit-organization', 'non-profit-organization']);
+        });
         if ($request->has('country')) {
             $country = $request->input('country');
             $query->where('country', $country);
--- a/app/Helpers/TerrafundDashboardQueryHelper.php
+++ b/app/Helpers/TerrafundDashboardQueryHelper.php
@@ -8,17 +8,16 @@ class TerrafundDashboardQueryHelper
 {
     public static function buildQueryFromRequest($request)
     {
-        $query = Project::query();
-        $query = $query->where('framework_key', 'terrafund');
-        $query = $query->whereHas('organisation', function ($query) {
-            $query->whereIn('type', ['for-profit-organization', 'non-profit-organization']);
-        });
+        $query = Project::where('framework_key', 'terrafund')
+            ->whereHas('organisation', function ($query) {
+                $query->whereIn('type', ['for-profit-organization', 'non-profit-organization']);
+            });
         if ($request->has('country')) {
             $country = $request->input('country');
-            $query->where('country', $country);
+            $query = $query->where('country', $country);
         } elseif ($request->has('uuid')) {
             $projectId = $request->input('uuid');
-            $query->where('uuid', $projectId);
+            $query = $query->where('uuid', $projectId);
         }
 
         return $query;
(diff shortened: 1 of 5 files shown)


app/Helpers/TerrafundDashboardQueryHelper.php
```diff
@@ -2,11 +2,17 @@
...
+use App\Models\V2\Projects\Project;
+
 class TerrafundDashboardQueryHelper
 {
-    public static function buildQueryFromRequest($query, $request)
+    public static function buildQueryFromRequest($request)
     {
+        $query = Project::query();
         $query = $query->where('framework_key', 'terrafund');
+        $query = $query->whereHas('organisation', function ($query) {
```
roguenet, line 13: A minor point, but there's no any real reason to assign the `$query` variable at each step
```
$query = Project::where('framework_key', 'terrafund')
    -> $whereHas('organisation', function($query) {
    ```

File: app/Helpers/TerrafundDashboardQueryHelper.php [hunk 2]
```diff
@@ -2,11 +2,17 @@
...
+            $query->whereIn('type', ['for-profit-organization', 'non-profit-organization']);
+        });
         if ($request->has('country')) {
             $country = $request->input('country');
             $query->where('country', $country);
```
roguenet, line 18: I see that this is older code, but is this functioning correctly? I had thought $query was immutable and you would need to save the return value. Maybe I'm misunderstanding how the query builder works, but I would think this is necessary:
```
$query = $query->where('country', $country);
```
(and same change in the `elseif` block)

app/Http/Controllers/V2/Dashboard/ActiveCountriesTableController.php
```diff
@@ -51,62 +52,50 @@ public function getAllCountries()
...
-        foreach ($projects as $project) {
-            $sites = $project->sites()->pluck('id')->toArray();
-            foreach ($sites as $siteId) {
-                $latestSiteReportId = SiteReport::where('site_id', $siteId)
-                    ->orderByDesc('due_at')
-                    ->value('id');
-                if ($latestSiteReportId !== null) {
-                    $totalSpeciesAmount += TreeSpecies::where('speciesable_id', $latestSiteReportId)->sum('amount');
+        return $projects->sum(function ($project) {
+            return $project->sites()->get()->sum(function ($site) {
```
roguenet, line 65: The `sum` function on `$projects` I think isn't working the way you want. That's a call on the query builder, which can't take a closure, only a [column name string](https://laravel.com/api/7.x/Illuminate/Database/Query/Builder.html#method_sum). Instead, you want `sum` on the collection, like you have on this line. So, I think for this to function correctly, it needs a `get()` in both cases.

```
return $projects->get()->sum(function($project) {
  return $project->sites()->get()->sum(function $site) {
```

However, that's running into N+ query land again, so a better approach would be to avoid the double loop:
```
return Sites::whereIn('project_id', $projects->pluck('id'))->get()->sum(function ($site) {
```

app/Http/Controllers/V2/Dashboard/ActiveCountriesTableController.php
```diff
@@ -51,62 +52,50 @@ public function getAllCountries()
...
+                $latestReport = $site->reports()->orderByDesc('due_at')->first();
+                if ($latestReport) {
+                    return $latestReport->treeSpecies()->sum('amount');
                 }
```
roguenet, line 69: Since this method is being implicitly typed to return an integer, this should likely `return 0;` after the if block for clarity.

app/Http/Controllers/V2/Dashboard/ActiveCountriesTableController.php
```diff
@@ -51,62 +52,50 @@ public function getAllCountries()
...
 
-    public function numberOfSites($country)
+    public function numberOfSites($country, $projects)
     {
-        $projects = $this->projectsCountry($country)->get();
+        $projects = $projects->where('country', $country);
 
         return $projects->sum(function ($project) {
             return $project->sites()->count();
         });
```
roguenet, line 93: This block can be two queries instead of N+1:
```
return Site::whereIn('project_id', $projects->pluck('id'))->count()
```

roguenet (commented):
This is getting better, but something I like to always keep in mind when writing big report endpoints like this is "how can I reduce the number of queries needed in this method / controller / etc". That's our biggest performance bottleneck

LimberHope:
Thanks for the correction, it makes sense, more than anything for good performance avoiding unnecessary methods or queries.

PR #65, opened 2024-03-05
[TM-706] Fix a couple of niggling bugs in the update request cycle.

roguenet:
https://gfw.atlassian.net/browse/TM-706


openapi-src/V2/paths/_index.yml
```diff
@@ -50,7 +50,9 @@
     $ref: './Forms/get-v2-admin-forms.yml'
 /v2/admin/reporting-frameworks:
   get:
-    $ref: './Forms/get-v2-admin-reporting-frameworks.yml'
+    $ref: './ReportingFrameworks/get-v2-admin-reporting-frameworks.yml'
+  post:
+    $ref: './ReportingFrameworks/post-v2-admin-reporting-frameworks.yml'
```
roguenet, line 55: I cannot figure out how the swagger doc on staging has this `post` path but it got removed on this epic branch. This reference is _not_ there in this file, but the built swagger doc has it. That leads me to believe there's something a little broken about the swagger doc generation process, but I'm not spending more time right now trying to figure it out. The FE build fails if this isn't here, and it's clearly supposed to be here since this is an actively used endpoint 🤷🏻

pachonjcl (approved):

PR #81, opened 2024-03-11
feat: TM-725 add new fields to v2 sites

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-725)

### DIFF
--- a/database/migrations/2024_03_11_132400_seed_options_list_siting_strategy.php
+++ b/database/migrations/2024_03_11_132400_seed_options_list_siting_strategy.php
@@ -1,9 +1,8 @@
 <?php
 
+use App\Helpers\I18nHelper;
 use App\Models\V2\Forms\FormOptionList;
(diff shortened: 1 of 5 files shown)


database/migrations/2024_03_11_132400_seed_options_list_siting_strategy.php
```diff
@@ -0,0 +1,58 @@
...
+                'status' => I18nItem::STATUS_DRAFT,
+                'short_value' => $short ? $value : null,
+                'long_value' => $short ? null : $value,
+            ]);
+
+            return $i18nItem->id;
+        }
+
+        return data_get($target, $property . '_id');
+    }
```
pachonjcl, line 57: to avoid using every time a new function called `generateIfMissingI18nItem` which is in multiple places, can you try to use the `I18nHelper`?

pachonjcl (changes requested):

pachonjcl (approved):

PR #107, opened 2024-03-27
[TM-748] re-instate ability to put site reports into the nothing to report state.

roguenet:
https://gfw.atlassian.net/browse/TM-748

This method was removed accidentally in an earlier PR.


pachonjcl (approved):

pachonjcl:
Just curious, where are you planning to use it?

roguenet:
> Just curious, where are you planning to use it?

It's already used (this method overrides the `false` from `HasReportStatus`, which uses the method to control whether a given report type can be put into a "nothing-to-report" state.

PR #135, opened 2024-04-09
[TM-785] Unit testing and linting for the BE repo

roguenet:
https://gfw.atlassian.net/browse/TM-785

* I removed all the legacy tests. They were slow, and most of them were testing dead code. They also did problematic things like potentially mess with the contents of our S3 buckets
* All other tests are fixed up so they run locally and on github actions. 
  * To test locally, run `make test`
* Linter and tests passing are now a requirement to merge into `staging`, `main` or `release/**`


app/Services/PushService.php
```diff
@@ -18,8 +18,12 @@ public function __construct()
         $this->snsClient = App::make('CustomSnsClient');
     }
 
-    public function fetchEndpointArn(string $os, string $pushToken): string
+    public function fetchEndpointArn(string $os, string $pushToken): ?string
     {
+        if (! config('app.sns.enabled')) {
+            return null;
+        }
```
roguenet, line 25: We should likely remove the SNS integration entirely as it's only useful for mobile apps, but for now this will do.

routes/api_v2.php
```diff
@@ -392,7 +390,7 @@
     Route::get('/projects', ViewMyProjectsController::class);
 });
 
-Route::post('/users/resend', [AuthController::class, 'resendByEmail']);
+Route::post('/users/resend', [AuthController::class, 'resendByEmail'])->withoutMiddleware('auth:service-api-key,api');
```
roguenet, line 393: This was the first bug found by the resurrected test suite. This endpoint currently doesn't work in production, but I as far as I know nobody has reported it yet.

pachonjcl (approved):
Lot of changes, glad that we have test back.

PR #142, opened 2024-04-13
[TM-812] Tree Species refactor

roguenet:
https://gfw.atlassian.net/browse/TM-812

FE: https://github.com/wri/wri-terramatch-website/pull/116

This turns tree species into a BE concern by avoiding separate mutation endpoints for tree species (create/update/delete). The problem there was that in the update request flow, the tree species on the base entity would be updated without admin approval of the update request.

I've also started sideloading the update request on the form entity API response so that the FE doesn't need to load both endpoints in order to load the form.


app/Models/Traits/UsesLinkedFields.php
```diff
@@ -208,24 +242,31 @@ private function updateLinkedFieldValue(array $linkedFieldInfo, $answer): void
         }
     }
 
-    public function getForm(): Form
+    private function syncRelation(string $property, $data): void
```
roguenet, line 245: This is the real meat of this PR: Over the course of this epic, we'll be adding support for several (hopefully all) of these relations so that the BE is responsible for managing them, rather than putting that onus on the FE, requiring it to access a bunch of extra APIs. This is both for performance and practical reasons (see the note about the update request flow in the PR description).

pachonjcl (approved):

PR #145, opened 2024-04-15
polygon validation endpoints

pachonjcl:
Polygon Validation work

Notes:

- Rebuild php container or make sure that the memory for upload and post are increased.

- make sure to create the following 3 folders inside `storage/app/public/` : [`geojson_files`,	`kml_files`,  `shapefiles`]

- Please make run the `database/seeders/CodeCriteria.php` and `database/seeders/CodeIndicator.php`

- [Documentation](https://gfw.atlassian.net/wiki/spaces/TerraMatch/pages/edit-v2/1048444932?draftShareId=c7a04889-09f5-4609-b9a2-6183e0fa2a63&inEditorTemplatesPanel=auto_closed)

image


app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -0,0 +1,757 @@
...
+    {
+        $validatedData = $request->validate([
+          'geometry' => 'required|json',
+        ]);
+        $geometry = json_decode($request->input('geometry'));
+        $geom = DB::raw("ST_GeomFromGeoJSON('" . json_encode($geometry) . "')");
+        $uuid = Str::uuid();
+
+        DB::table('polygon_geometry')->insert([
+          'uuid' => $uuid,
```
roguenet, line 44: UUID shouldn't be handled at the controller level, and neither should timestamps. If you include the `HasUuid` trait, you can drop them and it'll get generated on insert (`HasTimestamps` is included in the base `Model` class). Also, this would be a little more idiomatic if it weren't using the DB::table directly. This should do it:
```
$model = PolygonGeomemetry::create(['geom' => $geom]);
```

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -0,0 +1,757 @@
...
+          'geom' => $geom,
+          'created_at' => now(),
+          'updated_at' => now(),
+        ]);
+
+        return response()->json(['uuid' => $uuid], 200);
```
roguenet, line 50: With the change above, this will change to 
```
return response()->json(['uuid' => $model->uuid], 200);
```

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -0,0 +1,757 @@
...
+            $latitude = DB::selectOne("SELECT ST_Y(ST_Centroid(ST_GeomFromGeoJSON('$geojson'))) AS latitude")->latitude;
+            $areaSqMeters = $areaSqDegrees * pow(111320 * cos(deg2rad($latitude)), 2);
+
+            $areaHectares = $areaSqMeters / 10000;
+
+            $id = DB::table('polygon_geometry')->insertGetId([
+              'uuid' => $uuid,
+              'geom' => $geom,
+              'created_at' => now(),
+              'updated_at' => now(),
```
roguenet, line 72: Same guidance as above on `uuid`, and timestamps.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -0,0 +1,757 @@
...
+        }
+
+        return true;
+    }
+
+    public function validateDataInDB(Request $request)
+    {
+        $polygonUuid = $request->input('uuid');
+        $fieldsToValidate = ['poly_name', 'plantstart', 'plantend', 'practice', 'target_sys', 'distr', 'num_trees'];
+        $DATA_CRITERIA_ID = 14;
```
roguenet, line 154: What's the purpose of this magic number?

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -0,0 +1,757 @@
...
+        // Check if the polygon with the specified poly_id exists
+        $polygonExists = SitePolygon::where('poly_id', $polygonUuid)
+            ->exists();
+        $valid = false;
```
roguenet, line 158: This `$valid` variable is superfluous since it doesn't change any control flow logic. I'd just set the response in the block below to `'valid' => false` as well as in the other error block below, and finally return with `json(['valid' => true])`

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -0,0 +1,757 @@
...
+                $validSchema = false;
+                $validData = false;
+            } elseif (! $this->validateData($properties, $fieldsToValidate)) {
+                $validData = false;
+            }
+            $insertionSchemaSuccess = $this->insertCriteriaSite($polygonUuid, $SCHEMA_CRITERIA_ID, $validSchema);
+            $insertionDataSuccess = $this->insertCriteriaSite($polygonUuid, $DATA_CRITERIA_ID, $validData);
+
+            $sitePolygon = new SitePolygon();
+            $sitePolygon->uuid = Str::uuid();
```
roguenet, line 217: Same guidance as above on `uuid` and timestamps for this model.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -0,0 +1,757 @@
...
+        }
+
+        return $propertiesList;
+    }
+
+    public function uploadKMLFile(Request $request)
+    {
+        if ($request->hasFile('file')) {
+            $kmlfile = $request->file('file');
+            $directory = storage_path('app/public/kml_files');
```
roguenet, line 269: Maybe it's lower in the PR, but I'm going to drop a question here in case not: Is anything being done to clean up these files? I wouldn't want us to run out of space on the machines because we weren't monitoring this or cleaning it up over time. If they need to be kept long term, they should be getting sent to S3, rather than the local disk (nothing we need to keep long term should be stored on the EC2 machines; they should be considered ephemeral).

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -0,0 +1,757 @@
...
+    {
+        $uuid = $request->query('uuid');
+        $geometry = PolygonGeometry::where('uuid', $uuid)->first();
+
+        if (! $geometry) {
+            return response()->json(['error' => 'Geometry not found'], 404);
+        }
+
+        $isSimple = DB::selectOne('SELECT ST_IsSimple(geom) AS is_simple FROM polygon_geometry WHERE uuid = :uuid', ['uuid' => $uuid])->is_simple;
+        $SELF_CRITERIA_ID = 4;
```
roguenet, line 363: What's the purpose of this magic number?

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -0,0 +1,757 @@
...
+    //     $threshold = $averageDistance * 1.25; // Adjust this threshold as needed
+    //     for ($i = 0; $i < $numVertices - 1; $i++) {
+    //       $distance = $this->calculateDistance($coordinates[$i], $coordinates[$i + 1]);
+    //       if (abs($distance - $averageDistance) > $threshold) {
+    //         $spikes[] = $coordinates[$i];
+    //       }
+    //     }
+    //   }
+    //   return $spikes;
+    // }
```
roguenet, line 407: I'd like no commented out code to be committed to `staging`. Will this method be resurrected? I'm guessing not since there's another implementation below.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -0,0 +1,757 @@
...
+                    // Vertex and its adjacent vertices contribute more than 25% of the total boundary path distance
+                    $spikes[] = $coordinates[($i + 1) % $numVertices];
+                }
+            }
+        }
+
+        return $spikes;
+    }
+
+    public function insertCriteriaSite($POLYGON_ID, $CRITERIA_ID, $valid)
```
roguenet, line 436: all caps are typically reserved for constants. These should be `$polygonId` and `criteriaId`

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -0,0 +1,757 @@
...
+    {
+        $criteriaSite = new CriteriaSite();
+        $criteriaSite->polygon_id = $POLYGON_ID;
+        $criteriaSite->criteria_id = $CRITERIA_ID;
+        $criteriaSite->uuid = Str::uuid();
```
roguenet, line 441: Same comment as above with regards to UUIDs.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -0,0 +1,757 @@
...
+        $uuid = $request->query('uuid');
+        $geometry = PolygonGeometry::where('uuid', $uuid)->first();
+
+        if (! $geometry) {
+            return response()->json(['error' => 'Geometry not found'], 404);
+        }
+        $geojson = DB::selectOne('SELECT ST_AsGeoJSON(geom) AS geojson FROM polygon_geometry WHERE uuid = :uuid', ['uuid' => $uuid])->geojson;
+        $geojsonArray = json_decode($geojson, true);
+        $spikes = $this->detectSpikes($geojsonArray);
+        $SPIKE_CRITERIA_ID = 8;
```
roguenet, line 464: What's the purpose of this magic number?

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -0,0 +1,757 @@
...
+    {
+        $uuid = $request->query('uuid');
+        $geometry = PolygonGeometry::where('uuid', $uuid)->first();
+
+        if (! $geometry) {
+            return response()->json(['error' => 'Geometry not found'], 404);
+        }
+
+        $areaSqDegrees = DB::selectOne('SELECT ST_Area(geom) AS area FROM polygon_geometry WHERE uuid = :uuid', ['uuid' => $uuid])->area;
+        $latitude = DB::selectOne('SELECT ST_Y(ST_Centroid(geom)) AS latitude FROM polygon_geometry WHERE uuid = :uuid', ['uuid' => $uuid])->latitude;
```
roguenet, line 481: Not a big deal, but these two could be combined into a single query, which would be a little faster. Better yet would be to change these into accessors on the Eloquent model and do this instead:
```
$areaSqDegrees = $geometry->area_sq_degrees;
$latitude = $geometry->latitude;
```

Unfortunately, that route would go back to using two DB calls (although it would be easier to read). Maybe a single attribute on `PolygonGeometry` called `getDBGeometryAttribute` that calculates and returns both of these values in a single query.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -0,0 +1,757 @@
...
+        $areaSqMeters = $areaSqDegrees * pow(111320 * cos(deg2rad($latitude)), 2);
```
roguenet, line 482: What is this calculation doing? It looks potentially expensive

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -0,0 +1,757 @@
...
+        $SIZE_CRITERIA_ID = 6;
```
roguenet, line 483: What's the purpose of this magic number?

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -0,0 +1,757 @@
...
+        $geometry = PolygonGeometry::where('uuid', $polygonUuid)->first();
+
+        if (! $geometry) {
+            return response()->json(['error' => 'Geometry not found'], 404);
+        }
+
+        $totalArea = DB::table('polygon_geometry')
+            ->where('uuid', $polygonUuid)
+            ->selectRaw('ST_Area(geom) AS area')
+            ->first()->area;
```
roguenet, line 513: This could be an attribute on `PolygonGeometry` as well.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -0,0 +1,757 @@
...
+            ->first();
+
+        if (! $sitePolygonData) {
+            return response()->json(['error' => 'Site polygon data not found for the specified polygonUuid'], 404);
+        }
+
+        // Find the country ISO using project_id from v2projects
+        $countryIso = DB::table('v2_projects')
+            ->where('uuid', $sitePolygonData->project_id)
+            ->value('country');
```
roguenet, line 527: This can easily use the eloquent queries:
```
$countryIso = Project::isUuid($sitePolygonData->project_id)->select('county')->first()->country
```

Even better would be to establish a `belongsTo` relationship on `SitePolygon` -> `Project` (I don't even see `project_id` as a fillable on the model implementation?) so this could be:
```
$countryIso = $sitePolygonData->project()->select('country')->first()->country
```

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -0,0 +1,757 @@
...
+
+        if (! $countryIso) {
+            return response()->json(['error' => 'Country ISO not found for the specified project_id'], 404);
+        }
+
+        $intersectionData = DB::table('world_countries_generalized')
+          ->where('iso', $countryIso)
```
roguenet, line 534: Please use `WorldCountryGeneralized::where('iso', $countryIso)`

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -0,0 +1,757 @@
...
+          ->first();
+
+        $intersectionArea = $intersectionData->area;
+        $countryName = $intersectionData->country;
+
+        $insidePercentage = $intersectionArea / $totalArea * 100;
+
+        $insideThreshold = 75;
+        $insideViolation = $insidePercentage < $insideThreshold;
+        $WITHIN_COUNTRY_CRITERIA_ID = 7;
```
roguenet, line 545: I'll stop commenting on these now, but I'm super curious why:
* All these magic numbers are sprinkled around?
* If they really are necessary, they should be specified as constants in the controller, or if used in other code, in a shared location.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -0,0 +1,757 @@
...
+    {
+        $uuid = $request->input('uuid');
+
+        // Fetch the geometry type based on the UUID using SQL query
+        $query = 'SELECT ST_GeometryType(geom) AS geometry_type FROM polygon_geometry WHERE uuid = ?';
+        $result = DB::selectOne($query, [$uuid]);
+
+        if ($result) {
+            $geometryType = $result->geometry_type;
+            $valid = $geometryType === 'POLYGON' ? true : false;
```
roguenet, line 568: This can be simplified: `$valid = $geometryType === 'POLYGON';`

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -0,0 +1,757 @@
...
+        $criteriaList = [];
+        foreach ($criteriaData as $criteria) {
+            $criteriaId = $criteria->criteria_id;
+
+            // Check if the criteria is valid
+            $validCriteriaQuery = 'SELECT valid FROM criteria_site 
+                               WHERE polygon_id = ? AND criteria_id = ?';
+            $validResult = DB::selectOne($validCriteriaQuery, [$uuid, $criteriaId]);
+
+            $valid = $validResult ? $validResult->valid : null;
```
roguenet, line 612: This should use Eloquent:
```
$valid = CriteriaSite->where(['polygon_id => uuid, 'criteria_id' => $criteriaId])->select('valid')->first()?->valid;
```

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -0,0 +1,757 @@
...
+        }
+
+        return response()->json(['polygon_id' => $uuid, 'criteria_list' => $criteriaList]);
+    }
+
+    public function uploadGeoJSONFile(Request $request)
+    {
+        if ($request->hasFile('file')) {
+            $file = $request->file('file');
+            $directory = storage_path('app/public/geojson_files');
```
roguenet, line 628: Same concerns here as above with storing files on the EC2 machine.

roguenet (commented):
I don't see any updates to the API docs. Please add documentation in the `openapi-src` directory, and update the swagger doc with `make doc`.

I've only made it about halfway through the PR, but need to sign off for the day and wanted to get the comments so far to you. I'll continue reviewing tomorrow.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
pachonjcl: I uploaded a table on the description of the PR, that is the number referenced to.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
pachonjcl: I uploaded a table on the description of the PR, that is the number referenced to.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
pachonjcl: I uploaded a table on the description of the PR, that is the number referenced to.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
pachonjcl: 👍🏼

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
pachonjcl: suggestion implemented

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
pachonjcl: done.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
pachonjcl: Right now we haven't got any work on that. But yes, files should be discarded after processing.
We need to create a new ticket for that.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
pachonjcl: removed

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
pachonjcl: changed to camelCase

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
pachonjcl: changed.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
pachonjcl: changed.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
roguenet: I haven't dug in enough to tell for myself yet; does the table need to be in the DB? Is a join against these ever done at the SQL level? Or could this instead be in `config`, and the associate made at the code level?

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
roguenet: Resolving this as the conversation is continuing in an earlier comment.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
roguenet: Ok, please link here when the ticket is created.

app/Http/Controllers/V2/Terrafund/TerrafundEditGeometryController.php
```diff
@@ -0,0 +1,131 @@
...
+        $polygonGeometry = PolygonGeometry::where('uuid', $uuid)->first();
+        if (! $polygonGeometry) {
+            return response()->json(['message' => 'No polygon geometry found for the given UUID.'], 404);
+        }
+        $geojson = DB::table('polygon_geometry')
+        ->select(DB::raw('ST_AsGeoJSON(geom) as geojson'))
+        ->where('uuid', '=', $uuid)
+        ->get();
+
+        $geojsonData = json_decode($geojson[0]->geojson, true);
```
roguenet, line 55: This can be a bit cheaper in the DB and a little easier to read:
```
$geometryQuery = PolygonGeometry::isUuid($uuid);
if (! $gemoetryQuery->exists()) {
  return response()->json(['message' => 'No polygon geometry found for the given UUID.'], 404);
}

$geojsonData = json_decode($geometryQuery->select(DB::raw('ST_AsGeoJSON(geom) as geojson'))->first()->geojson, true));
```

docker/php.Dockerfile
```diff
@@ -33,3 +36,4 @@ RUN mv "$PHP_INI_DIR/php.ini-development" "$PHP_INI_DIR/php.ini"
 ## APACHE
 RUN a2enmod rewrite
 COPY docker/000-default.conf /etc/apache2/sites-available/000-default.conf
+COPY docker/php.ini /usr/local/etc/php/php.ini
```
roguenet, line 39: I'm surprised to see this change. Why was it needed?

roguenet (commented):
I've finished my initial read through. I think I can take care of it during my greenhouse API work, but a lot of the validation methods in the create geometry controller are going to need to be moved to a utility more accessible somewhere. Also, I think there might be room to tap into the Laravel Validator system for a bunch of these, cleaning up the readability and maintainability of them. I'll look into that as well.

I do wish we had unit tests for all this new functionality, but in the interest of time I'll withhold requesting that for now.

Once we've resolved the remaining outstanding comments, this should be ready to merge.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
egrojMonroy: updated the code to use Eloquent

app/Http/Controllers/V2/Terrafund/TerrafundEditGeometryController.php
egrojMonroy: improved code with the suggestion

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
pachonjcl: This will be used to join with other tables, just don't yet.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
pachonjcl: Here is the [ticket](https://gfw.atlassian.net/browse/TM-886)

docker/php.Dockerfile
pachonjcl: Mainly because we want to override some configurations for php.ini
Specifically we need to increase the file size limit.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
egrojMonroy: Move to getDBGeometryAttribute in model

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
egrojMonroy: Done the relation between sitePolygon and Project. 
Changed to suggested code

roguenet:
The linter needs some attention, and then I think this is just about ready.

roguenet (approved):

PR #146, opened 2024-04-15
[TM-833] Allow creation of new frameworks without breaking existing API access.

roguenet:
https://gfw.atlassian.net/browse/TM-833

I've also done a couple of other things:
* Consolidated the GH account creation stuff into the Roles Migration command, as that one should be used to maintain all permission we expect to be able to use in new code and tests
* Added a new `framework-terrafund-enterprises` permission in anticipation of the new framework mentioned in the ticket.
* Fixed a failing test from TM-782 (this sort of things won't make it into staging in the future, once all new branches have the unit testing fix on them).


app/Http/Controllers/Traits/IsAdminIndex.php
```diff
@@ -33,7 +33,7 @@ protected function isolateAuthorizedFrameworks(QueryBuilder $query, string $tabl
             $query->where(function ($query) use ($tableName, $frameworkNames, $user) {
                 foreach ($frameworkNames as $framework) {
                     $frameworkPermission = 'framework-' . $framework;
-                    if ($user->hasPermissionTo($frameworkPermission)) {
```
roguenet, line 36: `can` and `hasPermissionTo` appear to consistently test the same thing, but `can` doesn't simply returns false instead of throwing an error when the Permission hasn't been defined.

pachonjcl (approved):

PR #160, opened 2024-04-25
[TM-836] Merge entities tool

roguenet:
https://gfw.atlassian.net/browse/TM-836

For now it can only handle Terrafund Sites (and their reports), but it was built in a way that makes it easy to expand to other entities and frameworks in the future as needed.


roguenet:
Moved to Draft because Claire found some issues with the test run on staging.

pachonjcl (approved):

PR #189, opened 2024-05-08
[TM-849,848] feat: add swagger definitions and modify endpoints to create and uplo…

cesarLima1:
https://gfw.atlassian.net/browse/TM-849
https://gfw.atlassian.net/browse/TM-848


roguenet:
It looks like `swagger-v2.yml` was edited directly. Any change done that way will get blown away the next time someone runs `make doc`. Instead, please update the files in `openapi-src` and run `make doc`. That will probably fix the test failure on this branch as well.

PR #204, opened 2024-05-16
[TM-675] Include numerical Site/Project IDs in Site and Project API resources.

roguenet:
This involves:
* Dropping `old_model` from `v2_sites` and `v2_projects`
* Renaming `old_id` to `ppc_external_id` to make it more clear that this has a real purpose.
* Adding database triggers and constraints to make sure that PPC sites and projects have a unique numerical value in this field. A table in MySQL can't have more than one numerical auto incrementing key field, so adding these triggers was the next best thing to ensure things at the DB level, and it also meant that I could make sure it was only getting applied to PPC entities. We can't just use the `id`, because the PPC folks already have disseminated these `old_id`s far and wide, and it would be a really major disruption to change it at this point. I went back and forth with them a bit hoping they could switch to UUIDs, but it would just be too much lift on their end.

https://gfw.atlassian.net/browse/TM-675
[FE PR](https://github.com/wri/wri-terramatch-website/pull/186)


app/Http/Controllers/V2/Sites/CreateSiteWithFormController.php
```diff
@@ -23,14 +23,10 @@ public function __invoke(Form $form, CreateEntityFormRequest $formRequest)
...
         }
 
-        $lastOldId = Site::orderByDesc('old_id')
-            ->value('old_id');
-
         $site = Site::create([
             'framework_key' => $project->framework_key,
             'project_id' => $project->id,
             'status' => EntityStatusStateMachine::STARTED,
-            'old_id' => $lastOldId + 1,
```
roguenet, line 33: This has been replaced with  the DB triggers in the migration

pachonjcl (approved):
Nice work with the triggers.

PR #226, opened 2024-05-24
[TM-851] Modify Site Polygon table and change reference from est_area…

cesarLima1:
https://gfw.atlassian.net/browse/TM-851


database/migrations/2024_05_13_204255_modify_sites_polygons.php
```diff
@@ -0,0 +1,63 @@
...
+
+return new class () extends Migration {
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        if (! Schema::hasTable('site_polygon')) {
+            return;
+        }
```
roguenet, line 15: I suppose this check isn't hurting anything, but Laravel should guarantee that it always passes because the migrations get run in order.

roguenet (dismissed):

database/migrations/2024_05_13_204255_modify_sites_polygons.php
cesarLima1: This was just to ensure that someone running the migration independently wouldn't encounter issues. However, as you mentioned, there shouldn't be any problems with the execution since it proceeds in order.

cesarLima1:
@roguenet Just to confirm, after merging staging into this branch, can I proceed with merging this branch into staging and then delete it?

roguenet:
> @roguenet Just to confirm, after merging staging into this branch, can I proceed with merging this branch into staging and then delete it?

Yeah; if anything had been added that needed re-review, Github would automatically invalidate my approval. So if you merge in another branch and the approval doesn't go away, you're good to go and can merge.

cesarLima1:
@roguenet Could you check again? A commit was made to fix the merge conflict when merging from staging to this branch. In the code, `?string $site_id = null `was removed from the function `createGeojsonModels($geojson, $sitePolygonProperties = [])`. The correction was based on the change in this commit [81beb57](https://github.com/wri/wri-terramatch-api/pull/234/commits/81beb5741924cbb14820b3e9ba4222858726f746)

pachonjcl (approved):

PR #233, opened 2024-05-27
[TM-853] Add delete for polygon and site polygon

egrojMonroy:
https://gfw.atlassian.net/browse/TM-853


app/Helpers/GeometryHelper.php
```diff
@@ -33,4 +35,85 @@ public static function getPolygonsBbox($polygonsIds)
...
+            return null;
+        }
+
+        $sitePolygons = $project->sitePolygons;
+
+        if ($sitePolygons->isEmpty()) {
+            return null; // Return null if no polygons are found for the given projectUuid
+        }
+
+        $polyIds = $sitePolygons->pluck('poly_id')->toArray();
```
roguenet, line 89: I left this feedback on the other PR as well (is there duplicate code in these PRs? In the future, please base a second PR on the branch of the first so we can avoid reviewing code twice). This would be more efficient like so:
```
$polyIds = $project->sitePolygons()->pluck('poly_id')->toArray();
if (empty($polyIds)) {
  return null; // Return null if no polygons are found for the given projectUuid
}
```

app/Http/Controllers/V2/Terrafund/TerrafundEditGeometryController.php
```diff
@@ -28,6 +30,64 @@ public function getSitePolygonData(string $uuid)
...
+
+                    if ($centroid === null) {
+                        Log::warning("Invalid centroid for project UUID: $project->uuid");
+                    }
+                } else {
+                    Log::warning("Project with UUID $relatedSite->project_id not found.");
+                }
+            } else {
+                Log::warning("Site polygon with UUID $polygonGeometry->uuid not found.");
+            }
```
roguenet, line 54: This could avoid a lot of nesting by using early returns. Instead of 
```
if ($thing) {
  // do stuff
} else { 
  Log::warning(...)
}
```
use
```
if (!$thing) {
  Log::warning(...);
  return null;
}

// do stuff
```

egrojMonroy: Thanks, done for this code. I'll keep in mind for future uses.

app/Helpers/GeometryHelper.php
egrojMonroy: Done.

app/Http/Controllers/V2/Terrafund/TerrafundEditGeometryController.php
```diff
@@ -34,23 +34,22 @@ public function updateProjectCentroid($polygonGeometry)
     {
         try {
             $sitePolygon = SitePolygon::where('poly_id', $polygonGeometry->uuid)->first();
+            if (!$sitePolygon) {
+              Log::warning("Site polygon with UUID $polygonGeometry->uuid not found.");
+              return null;
+            }
+            $relatedSite = Site::where('uuid', $sitePolygon->site_id)->first();
+            $project = Project::where('id', $relatedSite->project_id)->first();
```
roguenet, line 42: SitePolygon has a relationship to Project through Site, so this whole thing can be a single query using built-in helpers:
```
$project = $sitePolygon->project
```

egrojMonroy: is changed to single query now.

roguenet (approved):

PR #234, opened 2024-05-27
[TM-848] add site id when uploading polygon and recalculate calc_area

cesarLima1:
https://gfw.atlassian.net/browse/TM-848

### DIFF
--- a/app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
+++ b/app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
@@ -60,14 +60,14 @@ class TerrafundCreateGeometryController extends Controller
     /**
      * @throws ValidationException
      */
-    public function insertGeojsonToDB(string $geojsonFilename)
+    public function insertGeojsonToDB(string $geojsonFilename, ?string $site_id = null)
     {
         $geojsonData = Storage::get("public/geojson_files/{$geojsonFilename}");
         $geojson = json_decode($geojsonData, true);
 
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -60,14 +60,14 @@ public function storeGeometry(Request $request)
...
-    public function insertGeojsonToDB(string $geojsonFilename)
+    public function insertGeojsonToDB(string $geojsonFilename, ?string $site_id = null)
     {
         $geojsonData = Storage::get("public/geojson_files/{$geojsonFilename}");
         $geojson = json_decode($geojsonData, true);
 
         SitePolygonValidator::validate('FEATURE_BOUNDS', $geojson, false);
 
-        return App::make(PolygonService::class)->createGeojsonModels($geojson);
+        return App::make(PolygonService::class)->createGeojsonModels($geojson, [], $site_id);
```
roguenet, line 70: I noticed this in the other PR, and was wondering how it was going to be used. This should instead be passed in as 
`createGeojsonModels($geojson, ['site_id' => $site_id])`
instead of requiring another parameter to the create method. That's how the greenhouse API endpoint specifies the site id.

cesarLima1: Thanks for the suggestion! It is implemented in commit [81beb57](https://github.com/wri/wri-terramatch-api/pull/234/commits/81beb5741924cbb14820b3e9ba4222858726f746)

roguenet (approved):

PR #240, opened 2024-05-29
[TM-671] return polygons, project and country data

egrojMonroy:
https://gfw.atlassian.net/browse/TM-671


app/Helpers/GeometryHelper.php
```diff
@@ -0,0 +1,119 @@
...
+            return null;
+        }
+
+        $sitePolygons = $project->sitePolygons;
+
+        if ($sitePolygons->isEmpty()) {
+            return null; // Return null if no polygons are found for the given projectUuid
+        }
+
+        $polyIds = $sitePolygons->pluck('poly_id')->toArray();
```
roguenet, line 25: This will be more efficient in the DB this way:
```
$polyIds = $project->sitePolygons()->pluck('poly_id')->toArray();
if ($polyIds->isEmpty()) {
  return null; // Return null if no polygons are found for the given projectUuid
}
```

Keep in mind that when accessing an association, if you access it without the function call (`->sitePolygons` vs `->sitePolygons()`), it will aggressively load all columns for every record. So if you only need a single column, accessing it this way is more efficient.

app/Helpers/GeometryHelper.php
```diff
@@ -0,0 +1,119 @@
...
+            Project::where('uuid', $projectUuid)
+                ->update([
+                    'lat' => $latitude,
+                    'long' => $longitude,
+                ]);
+
+
+            Log::info("Centroid updated for projectUuid: $projectUuid");
+
+            return 'Centroids updated successfully!';
```
roguenet, line 80: It's odd that this method is returning a JsonResponse in the `catch` block below and a string here. It should probably be a JsonResponse in both cases.

app/Helpers/GeometryHelper.php
```diff
@@ -0,0 +1,119 @@
...
+                $maxX = max($maxX, $x);
+                $minX = min($minX, $x);
+                $maxY = max($maxY, $y);
+                $minY = min($minY, $y);
+            }
+        }
+
+        $bboxCoordinates = [$minX, $minY, $maxX, $maxY];
+
+        return $bboxCoordinates;
```
roguenet, line 117: The `$bboxCoordinates` variable is superfluous - I'd simply return the array on line 115.

app/Http/Controllers/V2/Dashboard/CountryDataController.php
```diff
@@ -0,0 +1,110 @@
...
+        try {
+            $project = Project::isUuid($uuid)->first();
+
+            if (! $project) {
+                Log::error("Project not found for project with UUID: $uuid");
+            }
+            $countSitePolygons = 0;
+            if($project) {
+                $countSitePolygons = $project->getTotalSitePolygons();
+            }
```
roguenet, line 87: This can be simplified to 
```
$countSitePolygons = $project?->getTotalSitePolygons() ?? 0;
```

app/Http/Controllers/V2/Dashboard/CountryDataController.php
```diff
@@ -0,0 +1,110 @@
...
+            if (! $organization) {
+                Log::error("Organization not found for project with ID: $project->id");
+            }
+
+            $country = WorldCountryGeneralized::where('iso', $project->country)->first();
+            $data = [
+              ['key' => 'project_name', 'title' => 'title', 'value' => $project->name ?? null],
+              ['key' => 'country', 'title' => 'Country', 'value' => $country->country ?? null],
+              ['key' => 'polygon_counts', 'title' => 'No. of Site - Polygons', 'value' => $countSitePolygons ?? null],
+              ['key' => 'organizations', 'title' => 'Organization', 'value' => $organization->name ?? null],
```
roguenet, line 99: According to the checks above, project and organization could be null when constructing this data, so these need safe operators:
```
$country = $project ? WorldCountryGeneralized::where('iso', $project->country)->first() : null;
```
Access the records with `$country?->country`, `$project?->name` and `$organization?->name`

roguenet: Alternatively, exit the method early in the blocks above where an error is logged to avoid hitting this code with potentially null objects.

app/Models/V2/Projects/Project.php
```diff
@@ -359,17 +348,16 @@ public function getRegeneratedTreesCountAttribute(): int
...
+        $sumQueries = [
+            DB::raw('sum(`workdays_paid`) as paid'),
+            DB::raw('sum(`workdays_volunteer`) as volunteer'),
+        ];
+        $projectTotals = $this->reports()->hasBeenSubmitted()->get($sumQueries)->first();
+        // The groupBy is superfluous, but required because Laravel adds "v2_sites.project_id as laravel_through_key" to
+        // the SQL select.
+        $siteTotals = $this->submittedSiteReports()->groupBy('v2_sites.project_id')->get($sumQueries)->first();
+
+        return $projectTotals?->paid + $projectTotals?->volunteer + $siteTotals?->paid + $siteTotals?->volunteer;
```
roguenet, line 360: This change is reverting some recent work on `staging`. Please update your branch from `staging` to avoid that.

app/Models/V2/Projects/Project.php
```diff
@@ -181,8 +172,6 @@ class Project extends Model implements MediaModel, AuditableContract, EntityMode
         'restoration_strategy' => 'array',
         'sdgs_impacted' => 'array',
         'answers' => 'array',
-        'detailed_intervention_types' => 'array',
-        'states' => 'array',
```
roguenet, line 185: This removal and the one above is reverting some recent work from staging. Please update your branch to avoid that.

package.json
```diff
@@ -8,7 +8,8 @@
         "hot": "mix watch --hot",
         "prod": "npm run production",
         "production": "cross-env NODE_ENV=production node_modules/webpack/bin/webpack.js --no-progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
-        "doc-v2": "swagger-cli bundle openapi-src/v2.yml --outfile resources/docs/swagger-v2.yml --type yaml --dereference"
+        "doc-v2": "swagger-cli bundle openapi-src/v2.yml --outfile resources/docs/swagger-v2.yml --type yaml --dereference",
+        "doc-main": "swagger-cli bundle openapi-src/v2.yml --outfile resources/docs/swagger-v2.yml --type yaml --dereference"
```
roguenet, line 12: This looks identical to `doc-v2` - are both needed?

app/Helpers/GeometryHelper.php
egrojMonroy: Changed to JsonResponse on both

app/Helpers/GeometryHelper.php
egrojMonroy: Fixed

app/Http/Controllers/V2/Dashboard/CountryDataController.php
egrojMonroy: done

app/Http/Controllers/V2/Dashboard/CountryDataController.php
egrojMonroy: changed to exit if project dont exist and all access to records changed.

app/Models/V2/Projects/Project.php
egrojMonroy: done merge from staging

app/Models/V2/Projects/Project.php
egrojMonroy: done merge from staging

package.json
egrojMonroy: It is removed, because it wasnt needed.

app/Http/Controllers/V2/Dashboard/CountryDataController.php
```diff
@@ -80,21 +80,19 @@ public function getProjectData(string $uuid)
...
                 Log::error("Organization not found for project with ID: $project->id");
             }
 
-            $country = WorldCountryGeneralized::where('iso', $project->country)->first();
+            $country = $project ? WorldCountryGeneralized::where('iso', $project->country)->first() : null;
             $data = [
-              ['key' => 'project_name', 'title' => 'title', 'value' => $project->name ?? null],
-              ['key' => 'country', 'title' => 'Country', 'value' => $country->country ?? null],
+              ['key' => 'project_name', 'title' => 'title', 'value' => $project?->name ?? null],
+              ['key' => 'country', 'title' => 'Country', 'value' => $country?->country ?? null],
```
roguenet, line 95: Project is guaranteed not to be null here, so while the `$country?->` is still appropriate, line 92 doesn't need the ternary and line 94 should be `$project->name` instead of `$project?->name ?? null`. Also, `?? null` doesn't accomplish anything here, so this line can be simply `$country?->country`, which will evaluate to null if `$country` is null.

app/Models/V2/Projects/Project.php
```diff
@@ -124,6 +126,16 @@ class Project extends Model implements MediaModel, AuditableContract, EntityMode
...
+        'detailed_intervention_types',
+        'proj_impact_foodsec',
+        'pct_employees_marginalised',
+        'pct_beneficiaries_marginalised',
+        'pct_beneficiaries_men',
+        'proposed_gov_partners',
+        'proposed_num_nurseries',
+        'proj_boundary',
+        'states',
+        'proj_impact_biodiv',
```
roguenet, line 138: Not a huge deal because it'll even out in the end, but if the branch you're merging this into also got updated from staging, the diffs in this file should go away.

app/Models/V2/Projects/Project.php
```diff
@@ -496,4 +498,9 @@ private function submittedSiteReportIds(): array
     {
         return $this->submittedSiteReports()->pluck('v2_site_reports.id')->toArray();
     }
-}
+
+    public function getTotalSitePolygons()
+    {
+        return $this->sitePolygons()->count();
+    }
```
roguenet, line 505: This would be more idiomatic as an attribute:
```
public function getTotalSitePolygonsAttribute()
{
   return $this->sitePolygons()->count();
}
```

Then in the controller on line 85, you'd reference like this instead:
```
$countSitePolygons = $project->total_site_polygons;
```
Or just skip assigning it to a separate variable entirely and reference the attribute directly on line 96.

Finally, the `?? 0` on line 85 and `?? null` on line 96 in the controller and not needed. A `count()` call automatically returns 0 if there are no associated records.

app/Http/Controllers/V2/Dashboard/CountryDataController.php
```diff
@@ -0,0 +1,108 @@
...
+            if (! $organization) {
+                Log::error("Organization not found for project with ID: $project->id");
+            }
+
+            $country = $project ? WorldCountryGeneralized::where('iso', $project->country)->first() : null;
+            $data = [
+              ['key' => 'project_name', 'title' => 'title', 'value' => $project?->name ?? null],
+              ['key' => 'country', 'title' => 'Country', 'value' => $country?->country ?? null],
+              ['key' => 'polygon_counts', 'title' => 'No. of Site - Polygons', 'value' => $countSitePolygons ?? null],
+              ['key' => 'organizations', 'title' => 'Organization', 'value' => $organization->name ?? null],
```
roguenet, line 97: Since there's no early exit on a null organization above, this needs the safe navigation operator, and again here `?? null` isn't doing anything useful:
```
'value' => $organization?->name],
```

app/Http/Controllers/V2/Dashboard/CountryDataController.php
```diff
@@ -80,21 +80,19 @@ public function getProjectData(string $uuid)
 
             if (! $project) {
                 Log::error("Project not found for project with UUID: $uuid");
+                return response()->json(['error' => 'Project not found'], 404);
             }
-            $countSitePolygons = 0;
-            if($project) {
-                $countSitePolygons = $project->getTotalSitePolygons();
-            }
+            $countSitePolygons = $project?->getTotalSitePolygons() ?? 0;
```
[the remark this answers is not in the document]
egrojMonroy: understood, done

app/Models/V2/Projects/Project.php
egrojMonroy: removed and used directly the attribute accesor.

app/Http/Controllers/V2/Dashboard/CountryDataController.php
egrojMonroy: fixed

app/Models/V2/Projects/Project.php
egrojMonroy: removed

app/Http/Controllers/V2/Dashboard/CountryDataController.php
egrojMonroy: fixed

app/Models/V2/Projects/Project.php
```diff
@@ -131,11 +131,7 @@ class Project extends Model implements MediaModel, AuditableContract, EntityMode
         'pct_employees_marginalised',
         'pct_beneficiaries_marginalised',
         'pct_beneficiaries_men',
-        'proposed_gov_partners',
-        'proposed_num_nurseries',
-        'proj_boundary',
-        'states',
-        'proj_impact_biodiv',
```
roguenet, line 136: I'm confused why this is still showing that you're removing these lines. If this branch and the branch you're merging into are both updated from staging, there should be no diff here at all. These properties are needed, so I'm a bit worried that the changes you've made will cause them to get removed on `staging` once this is all merged in.

egrojMonroy: I got confused too with previous comment. But confirmed that these fields are on the target branch. Nonetheless to prevent deletion of these fields, I am returning them back, in this way the values will remain.

roguenet (approved):

PR #247, opened 2024-06-03
add endpoint to get polygons by status for given project

cesarLima1:
add endpoint to get polygons by status for given project

### DIFF
--- a/app/Helpers/TerrafundDashboardQueryHelper.php
+++ b/app/Helpers/TerrafundDashboardQueryHelper.php
@@ -36,7 +36,7 @@ class TerrafundDashboardQueryHelper
         return $polygonsIds;
     }
 
-    public static function getPolygonsByStatuses($projectUuid)
+    public static function retrievePolygonUuidsByStatusForProject($projectUuid)
     {
         $project = Project::where('uuid', $projectUuid)->first();
         $sitePolygons = $project->sitePolygons;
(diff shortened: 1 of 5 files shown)


app/Helpers/TerrafundDashboardQueryHelper.php
```diff
@@ -37,26 +36,36 @@ public static function getPolygonIdsOfProject($request)
...
+
+    public static function getPolygonsByStatusOfProject($request)
+    {
+        $projectUuid = TerrafundDashboardQueryHelper::buildQueryFromRequest($request)
+            ->pluck('uuid')->first();
+
+        return self::getPolygonsByStatuses($projectUuid);
+    }
+
+    public static function getPolygonsByStatus($request)
```
pachonjcl, line 65: I think this should be renamed to `getPolygonsByProjectStatus`

app/Helpers/TerrafundDashboardQueryHelper.php
```diff
@@ -37,26 +36,36 @@ public static function getPolygonIdsOfProject($request)
...
+                ->where('status', $status)
+                ->pluck('poly_id');
 
             $polygons[$status] = $polygonsOfProject;
         }
 
         return $polygons;
     }
+
+    public static function getPolygonsByStatusOfProject($request)
```
pachonjcl, line 57: This also has a very strange name. I suggest renaming it to `getPolygonsByProjectStatus`

app/Helpers/TerrafundDashboardQueryHelper.php
pachonjcl: please use better naming for the functions, right now I don't understand what are the difference between this two.

pachonjcl (changes requested):

pachonjcl (approved):

PR #251, opened 2024-06-04
[TM-920] Store PointGeometry in the GH API bulk upload geometry endpoint.

roguenet:
https://gfw.atlassian.net/browse/TM-920


routes/api_v2.php
```diff
@@ -552,10 +552,12 @@
     Route::get('/image/locations', SiteImageLocationsController::class);
     Route::delete('/', SoftDeleteSiteController::class);
     Route::get('/export', ExportAllSiteDataAsProjectDeveloperController::class);
+    // deprecated, use POST api/v2/geometry instead (include site_id in the geometry's properties
     Route::post('/geometry', [GeometryController::class, 'storeSiteGeometry']);
 });
```
cesarLima1, line 557: Just out of curiosity, will the deprecated endpoint `Route::post('/geometry', [GeometryController::class, 'storeSiteGeometry']);` be removed in the future?

cesarLima1 (approved):

routes/api_v2.php
roguenet: Yeah, I'd like to remove it as soon as I get confirmation from GH that they've switched to the other endpoint. Hopefully this won't be too disruptive for them since they're still in development on this API integration and it's not in their prod environment yet.

PR #254, opened 2024-06-06
[TM-857] Audit Log and Comment Feature for Project/Site/Polygons

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-857)

This PR is meant to add two tables Status and Attachment, which are going to be used for the audit log in Project, Sites and Polygons.

Two endpoints for `audit-status`

- Listing audit log for a given entity with entity uuid
- Saving a new audit log

One endpoint for 'site-polygon' which is going to be needed in the UI.

- List all polygons for a given project uuid (open for suggestion on what is the best way to use the paths)

One Trait that is going to be used later on to save entity status changes.

- SaveAuditStatusTrait

**Note:**
The next steps in this epic is to add the new status for Site and Polygon.
Make use of the `SaveAuditStatusTrait`
Add a couple of endpoints for updating status.


app/Http/Controllers/V2/AuditStatus/GetAuditStatusController.php
```diff
@@ -0,0 +1,84 @@
...
+use App\Http\Resources\V2\AuditStatusResource;
+use App\Models\V2\AuditStatus\AuditStatus;
+use App\Models\V2\Projects\Project;
+use App\Models\V2\Sites\Site;
+use App\Models\V2\Sites\SitePolygon;
+use Illuminate\Http\Request;
+
+class GetAuditStatusController extends Controller
+{
+    public function __invoke(Request $request, string $id = null)
```
roguenet, line 15: It looks like `$id` is not used below, and it doesn't appear to be part of the API definition. Please remove.

app/Http/Controllers/V2/AuditStatus/GetAuditStatusController.php
```diff
@@ -0,0 +1,84 @@
...
+    {
+        $auditStatus = AuditStatus::where('entity', $request->input('entity'))
+            ->where('entity_uuid', $request->input('uuid'))
```
roguenet, line 18: It looks like this is doing a polymorphic association, but it isn't using any of the tools we have in place to do these gracefully. 

This should be set up like so:
 * In `AuditAttachment`, the columns should be named something like `auditable_type` and `auditable_id`
 * A `morphTo` association should be created (see `Seeding.php` for a simple example)
 * The API endpoint should be structured with the entity type and ID in the URL (`/api/v2/audit-status/sites/{uuid}`)
 * With that done, the middleware for safely retrieving polymorphic associates can be used.
     * Create an interface (likely called `Auditable` - can be empty if there aren't any useful methods to include; I haven't looked deeply enough yet to tell. At the least, an association with auditStatuses would likely be helpful, like a `morphMany` accessor - see `SiteReport->seedings()`) and implement it on any model that AuditStatus can be attached to
     * In `ModelBindingInterfaceMiddleware`, make sure all of the classes that are Auditable are listed in `CONCRETE_MODELS` (from what I see here, likely just `SitePolygon` needs to be added)
     * The API definition for this route in `api_v2` then becomes something like
       ```
       Route::prefix('audit-status')->group(function () {
             ModelInterfaceBindingMiddleware::with(Auditable::class, function () {
                Route::get('/{auditable}', GetAuditStatusController::class);
            });
       ```
     * This method definition changes to
       ```
       public function __invoke(Request $request, Auditable $auditable) {
       ```

This method becomes smaller and more efficient (in combination with comments below; especially dropping the need for the N+1 `getEntity` query would be especially good)
```
$auditStatuses = auditable->auditStatuses()
  ->orderBy('updated_at', 'desc')
  ->orderBy('created_at', 'desc')
  ->with('auditAttachments')
  ->get();
return AuditStatusResource::collection($auditStatuses); 
```
return AuditStatusResource::collection(auditable->

app/Http/Controllers/V2/AuditStatus/GetAuditStatusController.php
```diff
@@ -0,0 +1,84 @@
...
+                'comment' => $audit->comment,
+                'date_created' => $audit->date_created,
+                'created_by' => $audit->created_by,
+                'type' => $audit->type,
+                'is_submitted' => $audit->is_submitted,
+                'is_active' => $audit->is_active,
+                'first_name' => $audit->first_name,
+                'last_name' => $audit->last_name,
+                'request_removed' => $audit->request_removed,
+                'attachments' => $this->mapAttachments($audit->auditAttachments),
```
roguenet, line 46: This work should all be done in `AuditStatusResource` in the `toArray` method, that's the purpose of that method in a resource. With the changes mentioned above, the the `'entity'` mapping would use (in the resource, `$this` maps to the underlying model being mapped): `$this->auditable->name`, and the association should already be mapping memory instead of requiring another query - `->auditable` will come from the `morphTo` association.

app/Http/Controllers/V2/AuditStatus/GetAuditStatusController.php
```diff
@@ -0,0 +1,84 @@
...
+    private function mapAttachments($attachments)
+    {
+        return $attachments->map(function ($attachment) {
+            return [
+                'id' => $attachment->id,
+                'entity_id' => $attachment->entity_id,
+                'file_name' => $attachment->file_name,
+                'file_url' => $attachment->file_url,
+                'date_created' => $attachment->date_created,
+                'created_by' => $attachment->created_by,
```
roguenet, line 60: This should be handled by `AttachmentResource`, and `AuditStatusResource`'s `attachments` association then becomes `'attachments' => AttachmentResource::collection($this->auditAttachments)`

app/Http/Controllers/V2/AuditStatus/StoreAuditStatusController.php
```diff
@@ -0,0 +1,67 @@
...
+use Illuminate\Support\Facades\Auth;
+use Illuminate\Support\Facades\Log;
+use Illuminate\Support\Facades\Storage;
+
+class StoreAuditStatusController extends Controller
+{
+    use SaveAuditStatusTrait;
+
+    public function __invoke(Request $request): AuditStatusResource
+    {
```
roguenet, line 20: This method will also benefit from the `ModelBindingInterfaceMiddelware` treatment mentioned above. With that, the `Auditable` will come in as a bound parameter to the method, as above. That means that `type` will no longer be sent in the body of the request, but will be in the URL along with the UUID.

Then this method gets a bit simpler as well. Note that I think the `type` member on `AuditStatus` becomes superfluous with these changes. Unless I'm missing something, the type of an audit is just the kind of model it's associated with.

This method should be able to be something more like:
```
  if ($auditable instanceof UpdateRequest) {
    $auditable->auditStatuses()->where('is_active', true)->update('is_active', false);
    $auditStatus = $this->saveAuditStatus($auditable, $body['status'], $body['comment'], $body['is_active'], $body['request_removed']; 
  } else {
    $auditStatus = $this->saveAuditStatus($auditable, $body['status'], $body['comment']);
  }
```

The file saving below is fine as-is I think, although see my comments on AuditAttachment; this would likely be better handled with Spatie. The `$auditStatusresponse` variable would be better named `$auditStatus` though, as that's what it is under the hood.

app/Models/V2/AuditAttachment/AuditAttachment.php
```diff
@@ -0,0 +1,29 @@
...
+
+class AuditAttachment extends Model
+{
+    use HasFactory;
+    use SoftDeletes;
+
+    public $table = 'attachment';
+
+    protected $fillable = [
+        'entity_id',
```
roguenet, line 18: This would be better named `audit_status_id` to reflect that it's attaching to.

routes/api_v2.php
```diff
@@ -653,6 +656,15 @@ function () {
 //Route::put('file/{uuid}', [FilePropertiesController::class, 'update']);
 //Route::delete('file/{uuid}', [FilePropertiesController::class, 'destroy']);
 
+Route::prefix('audit-status')->group(function () {
+    Route::get('/', GetAuditStatusController::class);
+    Route::post('/', StoreAuditStatusController::class);
+});
+
+Route::prefix('site-polygon')->group(function () {
+    Route::get('/project/{uuid}', GetPolygonByProjectController::class);
```
roguenet, line 665: I'm confused why this endpoint is needed when we already have `/api/v2/projects/{uuid}/site-polygons` defined?

app/Models/V2/AuditAttachment/AuditAttachment.php
```diff
@@ -0,0 +1,29 @@
...
+        'file_name',
+        'file_url',
+        'date_created',
+        'created_by',
+    ];
+
+    public function auditStatus()
+    {
+        return $this->belongsTo(AuditStatus::class, 'entity_id', 'id');
+    }
```
roguenet, line 28: As I've read more of the PR, I think this table is unneeded. We already use Spatie for attaching media of any type to a model, so this should use that system instead. Please see the `HasV2MediaCollection` and its uses as an example. `Organisation` in particular has lots of "media" that are files like spreadsheets, so it's not just images.

app/Http/Controllers/V2/AuditStatus/GetAuditStatusController.php
```diff
@@ -0,0 +1,21 @@
...
+namespace App\Http\Controllers\V2\AuditStatus;
+
+use App\Http\Controllers\Controller;
+use App\Http\Resources\V2\AuditStatusResource;
+use App\Models\V2\EntityModel;
+use Illuminate\Http\Request;
+
+class GetAuditStatusController extends Controller
+{
+    public function __invoke(Request $request, EntityModel $entity)
```
roguenet, line 12: Are all EntityModels auditable? I'd been proposing having a new interface to limit what can be passed into this controller to just those types of models that can be audited. Also, it looks from other parts of this PR that `SitePolygon` can be audited, and it doesn't implement `EntityModel`.

app/Http/Controllers/V2/AuditStatus/StoreAuditStatusController.php
pachonjcl: There is another `Auditable` already, so I went for `AuditableModel` following `EntityModel`.
I just couldn't make the store method to work with `ModelInterfaceBindingMiddleware`.

But I think this is good enough improvement.

app/Http/Controllers/V2/AuditStatus/GetAuditStatusController.php
```diff
@@ -4,18 +4,24 @@
...
+        $auditStatuses = $auditable->auditStatuses()
             ->orderBy('updated_at', 'desc')
             ->orderBy('created_at', 'desc')
             ->get();
 
+        $auditStatuses->map(function ($auditStatus) use ($auditable) {
+            $auditStatus->entity_name = $auditable->name;
+
+            return $auditStatus;
+        });
```
roguenet, line 23: It's weird to run a `map` function and not do anything with the result. This is probably a bit cleaner as a simple foreach. That also avoids the potential problem of the assignment not working because `$auditStatus` isn't being passed by reference
```
foreach ($auditStatuses as $auditStatus) {
  $auditStatus->entity_name = $auditable->name;
}
```

app/Http/Controllers/V2/AuditStatus/StoreAuditStatusController.php
```diff
@@ -4,82 +4,47 @@
...
-            'file_name' => $file->getClientOriginalName(),
-            'file_url' => $urlPath,
-            'date_created' => now(),
-            'created_by' => Auth::user()->email_address,
-        ]);
+        return new AuditStatusResource($auditStatus);
     }
 
-    private function getAuditableTypeFromRequest(Request $request)
+    private function getEntityFromRequest(AuditStatusCreateRequest $request)
```
roguenet, line 39: This method should not be needed if the `ModelBindingInterfaceMiddleware` is used for this endpoint as well.

app/Models/V2/AuditableModel.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace App\Models\V2;
+
+interface AuditableModel
+{
+}
```
roguenet, line 7: Since `auditStatuses` is being referenced on this interface in `GetPolygonByProjectController`, it should be defined here as well. I sure wish PHP were just a little more type safe 🤷🏻

app/Policies/V2/AuditStatus/AuditStatusPolicy.php
```diff
@@ -0,0 +1,15 @@
...
+
+use App\Models\User;
+use App\Models\V2\AuditStatus\AuditStatus;
+use App\Policies\Policy;
+
+class AuditStatusPolicy extends Policy
+{
+    public function uploadFiles(?User $user, ?AuditStatus $auditStatus): bool
+    {
+        return true;
```
roguenet, line 13: There isn't much point in a policy that is always true. Is this really valid? If it is, I think simply removing this policy and the the policy check in the controller that uses it would be fine.

database/migrations/2024_06_06_023733_create_status_table.php
```diff
@@ -0,0 +1,41 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration {
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::dropIfExists('status');
```
roguenet, line 13: `status` is an awfully generic table name, and breaks the class <-> table naming conventions in laravel. I think if you named this `audit_statuses` instead, it would be more intuitive, and you'd be able to drop the `$table` member on `AuditStatus`, because the name would match what laravel expects.

roguenet (commented):
One last minor request, otherwise looking good.

roguenet (approved):

roguenet (dismissed):

roguenet (approved):

roguenet (approved):

PR #262, opened 2024-06-07
[TM-967] feat: add project profile information controller

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-967)


app/Http/Controllers/V2/Dashboard/ProjectProfileDetailsController.php
```diff
@@ -0,0 +1,38 @@
...
+                'country' => $this->getCountry($project->country),
+                'countrySlug' => $project->country,
+                'organisation' => $project->organisation->type,
+                'survivalRate' => $project->survival_rate,
+                'restorationStrategy' => $project->restoration_strategy,
+                'targetLandUse' => $project->land_use_types,
+                'landTenure' => $project->land_tenure_project_area,
+            ];
+
+            return new ProjectProfileDetailsResource($response);
```
roguenet, line 29: Same comment here as on the others; please drop the resource when it's not doing any work other than mapping one array onto another with the same keys.

app/Http/Controllers/V2/Dashboard/ProjectProfileDetailsController.php
```diff
@@ -0,0 +1,38 @@
...
+use App\Http\Resources\V2\Dashboard\ProjectProfileDetailsResource;
+use App\Models\V2\Forms\FormOptionListOption;
+use App\Models\V2\Projects\Project;
+use Illuminate\Http\Request;
+
+class ProjectProfileDetailsController extends Controller
+{
+    public function __invoke(Request $request): ProjectProfileDetailsResource
+    {
+        if ($request->has('uuid')) {
```
roguenet, line 15: Since `uuid` is required to do anything useful in this endpoint, please include it in the URL instead of the query. The `api_v2` definition should be:
```
Route::get('/project-details/{project}', ProjectProfileDetailsController::class)
```

And then this method becomes 
```
public function __invoke(Request $request, Project $project): JsonResource
```

Laravel will know that you're seeking a Project instance with the UUID in the URL, get it out of the database and pass it directly into the method.

app/Http/Controllers/V2/Dashboard/ProjectProfileDetailsController.php
```diff
@@ -3,32 +3,28 @@
...
 use App\Http\Controllers\Controller;
-use App\Http\Resources\V2\Dashboard\ProjectProfileDetailsResource;
 use App\Models\V2\Forms\FormOptionListOption;
 use App\Models\V2\Projects\Project;
 use Illuminate\Http\Request;
 
 class ProjectProfileDetailsController extends Controller
 {
-    public function __invoke(Request $request): ProjectProfileDetailsResource
+    public function __invoke(Request $request, string $project = null)
```
roguenet, line 12: `This should be `Project $project` instead of `string $project = null`. And then fetching the project on line 14 goes away.

roguenet (approved):

PR #265, opened 2024-06-07
[TM-844] feat: add project pipeline controllers

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-844)


roguenet:
Please resolve conflicts with the base branch. I'll review once those are sorted out.

app/Http/Requests/V2/UpdateProjectPipelineRequest.php
```diff
@@ -0,0 +1,23 @@
...
+        return [
+            'Name' => ['sometimes', 'string', 'nullable', 'max:256'],
+            'SubmittedBy' => ['sometimes', 'string', 'nullable', 'max:256'],
+            'Description' => ['sometimes', 'string', 'nullable', 'max:500'],
+            'Program' => ['sometimes', 'string', 'nullable', 'max:256'],
+            'Cohort' => ['sometimes', 'string', 'nullable', 'max:256'],
+            'PublishFor' => ['sometimes', 'string', 'nullable', 'max:256'],
+            'URL' => ['sometimes', 'string', 'nullable', 'max:256'],
+            'CreatedDate' => ['sometimes', 'nullable', 'date'],
+            'ModifiedDate' => ['sometimes', 'nullable', 'date'],
```
roguenet, line 20: It looks like the rules for the store and update endpoints use the same rules (which makes sense). These should get combined into a single `ProjectPipelineRequest` object used on both endpoints.

app/Models/V2/ProjectPipeline.php
```diff
@@ -0,0 +1,32 @@
...
+    protected $fillable = [
+        'Name',
+        'SubmittedBy',
+        'Description',
+        'Program',
+        'Cohort',
+        'PublishFor',
+        'URL',
+        'CreatedDate',
+        'ModifiedDate',
```
roguenet, line 25: Please move these columns to snake_case, as is the norm for all tables in our DB.

roguenet: I'm also surprised that I don't see a migration to create the table in this PR. Has it already been merged? Typically the creation of the model .php and the migration to create the DB table would be part of the same PR.

roguenet: Also, it's odd to me that this would have custom `CreatedDate` and `ModifiedDate` columns when almost all of our models use the built-in Laravel `created_at` and `updated_at`, which get set automatically.

roguenet:
Please resolve conflicts with the base branch.

database/migrations/2024_06_07_163000_create_project_pipeline_table.php
```diff
@@ -0,0 +1,28 @@
...
+    public function up()
+    {
+        Schema::create('project_pipeline', function (Blueprint $table) {
+            $table->id();
+            $table->string('name', 256)->nullable();
+            $table->string('submitted_by', 256)->nullable();
+            $table->string('description', 500)->nullable();
+            $table->string('program', 256)->nullable();
+            $table->string('cohort', 256)->nullable();
+            $table->string('publish_for', 256)->nullable();
```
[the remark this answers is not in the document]
LimberHope: Yes, the purpose of the `submitted_by` field is to store the user's name for the moment. Would it be relevant to associate it as a foreign key?

The `publish_for` field is to store a URL.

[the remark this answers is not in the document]
roguenet: Yeah, it would be better to associate it to the user record and pull the name from there when needed in an API response. Storing the user's name directly is pretty fragile - what if the user record name changes?

roguenet (approved):

PR #270, opened 2024-06-07
[TM-970] Stop trying to call a script that doesn't exist.

roguenet:
https://gfw.atlassian.net/browse/TM-970


pachonjcl (approved):
Yeah I saw some error logs in Sentry pointing to this non existing job.
Nice catch

PR #275, opened 2024-06-10
[TM-941] Audit Log request support

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-941)

Main focus of this PR is to add a update status method for  `AuditableModel` .
Right now we are restricting the approval in `Site` and `SitePolygon` based on some conditions.

I also moved some `CriteriaSite` validation to `GeometryHelper` so I can reuse the code.

### DIFF
--- a/app/Helpers/GeometryHelper.php
+++ b/app/Helpers/GeometryHelper.php
@@ -5,7 +5,6 @@ namespace App\Helpers;
 use App\Models\V2\PolygonGeometry;
 use App\Models\V2\Projects\Project;
 use App\Models\V2\Sites\CriteriaSite;
-use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Log;
 
 class GeometryHelper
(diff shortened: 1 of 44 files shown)


app/Helpers/GeometryHelper.php
```diff
@@ -117,4 +119,29 @@ public static function getPolygonsBbox($polygonsIds)
...
+
+    public static function getCriteriaDataForPolygonGeometry($polygonGeometry)
+    {
+        // Fetch data from criteria_site with distinct criteria_id based on the latest created_at
+        $criteriaDataQuery = 'SELECT criteria_id, MAX(created_at) AS latest_created_at
+                            FROM criteria_site 
+                            WHERE polygon_id = ?
+                            GROUP BY criteria_id';
+
+        $criteriaData = DB::select($criteriaDataQuery, [$polygonGeometry->uuid]);
```
roguenet, line 131: Please lean into the Laravel associations as much as possible. For one thing, this query as written is completely ignoring the soft deletes feature. This should get you what you want instead:
```
$polygonGeometry->criteriaSite()->selectRaw('criteria_id, max(created_at) as latest_created_at')->groupBy('criteria_id')->get();
```

app/Helpers/GeometryHelper.php
```diff
@@ -117,4 +119,29 @@ public static function getPolygonsBbox($polygonsIds)
...
+
+        // Determine the validity of each criteria
+        $criteriaList = [];
+        foreach ($criteriaData as $criteria) {
+            $criteriaId = $criteria->criteria_id;
+            $valid = CriteriaSite::where(['polygon_id' => $polygonGeometry->uuid, 'criteria_id' => $criteriaId])->orderBy('created_at', 'desc')->select('valid')->first()?->valid;
```
roguenet, line 137: Likewise this should use the laravel assocations
```
$valid = $polygonGeometry->criteriaSite()->where('criteria_id', $criteriaId)->orderby('created_at', 'desc')->select('valid')->first()?->valid;
```
Even better would be to add a scope to CriteriaSite like this:
```
    public function scopeIsCriteria($query, $criteriaId): Builder
    {
        return $query->where('criteria_id', $criteriaId);
    }
 ```
 
so that this can become:
```
$valid = $polygonGeometry->criteriaSite()->isCriteria($criteriaId)->orderBy('created_at', 'desc')->select('valid')->first()?->valid;
```

What I'd really like to see is a way to replace this N+1 queries method with a single query, but after playing around with MariaDB group and distinct queries for awhile, I think there just isn't a super reasonable way to do it.

Comment (roguenet) [reply app/Helpers/GeometryHelper.php:137]: Ah, I realized there is a way to do it. Since max(created_at) and max(id) should always result in the same row, we can do this:
```
CriteriaSite::whereIn('id', $pg->criteriaSite()->groupBy('criteria_id')->selectRaw('max(id) as latest_id'))->get();
```
This gets all the data for this method in one query. The resulting SQL looks good to me:
```
      "query" => "select * from `criteria_site` where `id` in (select max(id) as latest_id from `criteria_site` where `criteria_site`.`polygon_id` = ? and `criteria_site`.`polygon_id` is not null and `criteria_site`.`deleted_at` is null group by `criteria_id`) and `criteria_site`.`deleted_at` is null",
      "bindings" => [
        "e7b564e6-3590-4215-a0eb-892a607f2075",
      ],
```

app/Helpers/GeometryHelper.php
pachonjcl: really great improvement.
implemented

app/Helpers/GeometryHelper.php
pachonjcl: really great improvement.
implemented

roguenet (approved):

PR #277, opened 2024-06-11
[TM-851] convert to post request for polygon validation

cesarLima1:
https://gfw.atlassian.net/browse/TM-851
EPIC: https://gfw.atlassian.net/browse/TM-853

### DIFF
--- a/openapi-src/V2/paths/Terrafund/get-v2-terrafund-validation-polygon.yml
+++ b/openapi-src/V2/paths/Terrafund/get-v2-terrafund-validation-polygon.yml
@@ -1,4 +1,4 @@
-summary: Get validation results for a polygon
+summary: Run validation for a polygon
 parameters:
   - in: query
     name: uuid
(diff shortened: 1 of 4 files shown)


app/Services/PolygonService.php
```diff
@@ -83,7 +83,6 @@ public function createCriteriaSite($polygonId, $criteriaId, $valid): bool|string
         $criteriaSite->polygon_id = $polygonId;
         $criteriaSite->criteria_id = $criteriaId;
         $criteriaSite->valid = $valid;
-        $criteriaSite->created_by = Auth::user()?->id;
```
roguenet, line 86: Why was this removed?

cesarLima1: It was an error when testing some features in our enviroment. criteria_site doesn't have that column.

roguenet: Oh 🤦🏻 😂

PR #289, opened 2024-06-17
Add new roles to users

LimberHope:
add new roles for users, project_developer, funder, government and new fields (country, program) to the users table
[Task](https://gfw.atlassian.net/browse/TM-599)


app/Console/Commands/Migration/RolesMigrationCommand.php
```diff
@@ -28,7 +28,7 @@ class RolesMigrationCommand extends Command
     public function handle()
     {
         if ($this->option('log')) {
-            echo('* * * Started * * * ' . $this->description . chr(10));
+            echo ('* * * Started * * * ' . $this->description . chr(10));
```
roguenet, line 31: The linter hasn't completed as of this review, bug I suspect it'll complain about this change. I don't see why this space is needed.

app/Console/Commands/Migration/RolesMigrationCommand.php
```diff
@@ -88,7 +103,7 @@ public function handle()
             });
 
         if ($this->option('log')) {
-            echo('- - - Finished - - - ' . chr(10));
+            echo ('- - - Finished - - - ' . chr(10));
```
roguenet, line 106: same as above with this change.

app/functions.php
```diff
@@ -154,7 +154,7 @@ function assignSpatieRole($user)
 {
     switch($user->role) {
         case 'user':
-            $user->assignRole('project-developer');
+            $user->assignRole('project_developer');
```
roguenet, line 157: Is this change intentional? This seems like it might have some pretty serious downstream consequences; changing the name of this role.

app/Console/Commands/Migration/RolesMigrationCommand.php
pachonjcl: space removed

app/Console/Commands/Migration/RolesMigrationCommand.php
pachonjcl: same

app/functions.php
pachonjcl: changing it back to use dash instead of underscore

app/Policies/Policy.php
```diff
@@ -37,6 +37,13 @@ protected function isAdmin(?UserModel $user): bool
         return ! $this->isGuest($user) && $user->role == 'admin';
     }
 
+    protected function isNewRoleUser(?UserModel $user): bool
+    {
+        $newRoles = ['project_developer', 'funder', 'government', 'user'];
```
roguenet, line 42: Should this be `project-developer`?

pachonjcl: yes, changed.

roguenet (dismissed):

roguenet (approved):

PR #306, opened 2024-06-25
[TM-855] feat: add endpoint to change status polygon

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-855)


app/Http/Controllers/V2/Polygons/ChangeStatusPolygonsController.php
```diff
@@ -0,0 +1,32 @@
...
+class ChangeStatusPolygonsController extends Controller
+{
+    use SaveAuditStatusTrait;
+
+    public function __invoke(ChangeStatusPolygonsUpdateRequest $request)
+    {
+        $body = $request->all();
+        $polygons = $body['updatePolygons'];
+
+        foreach ($polygons as $polygon) {
```
roguenet, line 19: Any time I see DB queries being executed in a loop, I want that N+1 query state to be improved. In this case, I think you should be able to do something more like
```
$sitePolygons = SitePolyon::whereIn('uuid', collect($polygons)->map(fn ($p) => $p['uuid'])->get();
foreach ($sitePolygons as $sitePolygon) { 
  ....
```

You might have to do a little extra to find the associated polygon status from the $polygons array, but that's better than issuing a DB query per polygon

pachonjcl: changed

app/Http/Controllers/V2/Polygons/ChangeStatusPolygonsController.php
```diff
@@ -14,19 +14,22 @@ class ChangeStatusPolygonsController extends Controller
...
-        $polygons = $body['updatePolygons'];
+        $updatePolygons = $body['updatePolygons'];
 
-        foreach ($polygons as $polygon) {
-            $sitePolygon = SitePolygon::where('uuid', $polygon['uuid'])->first();
-            if (! $sitePolygon) {
+        $sitePolygons = SitePolygon::whereIn('uuid', collect($updatePolygons)->map(fn ($p) => $p['uuid']))->get();
+        $changedPolygons = [];
+        foreach ($sitePolygons as $sitePolygon) {
+            $foundPolygon = collect($updatePolygons)->first(fn ($p) => $p['uuid'] === $sitePolygon->uuid);
```
roguenet, line 22: You're collecting the polygons many times. Would be better to do it once on line 17 and use the collection everywhere else.

roguenet (approved):

PR #308, opened 2024-06-26
[TM-1013] all audit logs in nurseries report pages

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-1013)
join information from audits and audit_statuses tables


app/Http/Controllers/V2/AuditStatus/GetAuditStatusController.php
```diff
@@ -20,6 +26,92 @@ public function __invoke(Request $request, AuditableModel $auditable)
...
+        $transformedAudits = $this->transformAudits($this->getAudits($auditable), $auditable);
+
+        $combinedData = $auditStatuses->concat($transformedAudits);
+
+        return AuditStatusResource::collection($combinedData);
+    }
+
+    private function getAudits($auditable)
+    {
+        $auditsModelInstance = $this->getModelInstance($auditable);
```
roguenet, line 38: I don't understand what's going on with this `getModelInstance` method. `$auditable` is already a model instance, and that method appears to be returning a model *class*, not instance, which means the `->audits()` call below doesn't look like it would function at all. Can you enlighten me as to how this works? At the very least, I don't think the `getModelInstance()` method should be needed at all; avoiding methods like that is the whole purpose of the ModelBindingInterfaceMiddleware, which this endpoint is using.

app/Http/Controllers/V2/AuditStatus/GetAuditStatusController.php
```diff
@@ -20,6 +26,92 @@ public function __invoke(Request $request, AuditableModel $auditable)
...
+                'status' => $audit->new_values['status'] ?? null,
+                'comment' => $audit->new_values['feedback'] ?? null,
+                'first_name' => $audit->user->first_name ?? null,
+                'last_name' => $audit->user->last_name ?? null,
+                'type' => 'status',
+                'is_submitted' => $audit->is_submitted ?? null,
+                'is_active' => $audit->is_active ?? null,
+                'request_removed' => $audit->request_removed ?? null,
+                'date_created' => $audit->created_at ?? null,
+                'created_by' => $audit->user_id ?? null,
```
roguenet, line 67: This kind of mapping should be handled in the resource (in this case, `AuditStatusResource`; this mapping shouldn't be needed in the controller.

roguenet (approved):

PR #309, opened 2024-06-27
[TM-999] integrate point endpoint to terramach

egrojMonroy:
https://gfw.atlassian.net/browse/TM-999


app/Services/PolygonService.php
```diff
@@ -212,24 +215,25 @@ protected function validateSitePolygonProperties(string $polygonUuid, array $pro
...
      */
-    protected function transformAndStorePoints($geojson, $sitePolygonProperties): string
+    protected function transformAndStorePoints($geojson, $sitePolygonProperties): array
     {
-        $pointUuids = [];
-        foreach ($geojson['features'] as $feature) {
-            $pointUuids[] = $this->insertSinglePoint($feature);
+        foreach ($geojson['features'] as &$feature) {
+            $currentPointUUID = $this->insertSinglePoint($feature);
+            $feature['properties']['point_id'] = $currentPointUUID;
```
roguenet, line 222: Is this assignment working? I don't understand why, but I've had trouble with `&$` in loops and had to resort to using `data_set` instead. But if you've tested and this is working as expected, I'm happy 👍🏻

app/Services/PolygonService.php
```diff
@@ -212,24 +215,25 @@ protected function validateSitePolygonProperties(string $polygonUuid, array $pro
...
+            }
         }
 
-        // TODO:
-        //  * transform points into a polygon
-        //  * Insert the polygon into PolygonGeometry
-        //  * Create the SitePolygon using the data in $properties (including $properties['site_id'] to identify the site)
-        //  * Return the PolygonGeometry's real UUID instead of this fake return
+        $polygonsGeojson = App::make(PythonService::class)->voronoiTransformation($geojson);
+        $polygonsUuids = $this->createGeojsonModels($polygonsGeojson, ['site_id' => $mainSiteID]);
```
roguenet, line 235: You should be able to get this from `$properties['site_id']` instead of assigning it to a custom variable in the loop.

Also, I believe the `properties` should be assigned to all of the site polygons, not just the ones they happen to be set on; which is why that loop is there to begin with. I think this line should simply be (note I'm returning here, as the `polygonsUuids` variable is superfluous):
```
return $this->createGeojsonModels($polygonsGeojson, $properties);
```

app/Services/PolygonService.php
```diff
@@ -212,24 +215,25 @@ protected function validateSitePolygonProperties(string $polygonUuid, array $pro
...
+                $mainSiteID = $properties[$property];
+            }
         }
 
-        // TODO:
-        //  * transform points into a polygon
-        //  * Insert the polygon into PolygonGeometry
-        //  * Create the SitePolygon using the data in $properties (including $properties['site_id'] to identify the site)
-        //  * Return the PolygonGeometry's real UUID instead of this fake return
+        $polygonsGeojson = App::make(PythonService::class)->voronoiTransformation($geojson);
```
roguenet, line 234: One more thing; please throw an exception if `$polygonsGeojson` is null. That's causing an error in the unit test, but is a legit problem that could happen in prod as well if something goes wrong. I'm working on fixing up the unit test now.

app/Services/PolygonService.php
egrojMonroy: Yes, it is working fine in local. Hope it works everywhere.

app/Services/PolygonService.php
egrojMonroy: Now it throws the exception now.

app/Services/PolygonService.php
egrojMonroy: Here is now only using properties that come from point, due to the change in understanding that everypoint will have its own attributes for site polygon.

roguenet (approved):

PR #321, opened 2024-07-02
[TM-1025] add isNewRoleUser to me function

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-1025)

### DIFF
--- a/app/functions.php
+++ b/app/functions.php
@@ -170,6 +170,9 @@ function assignSpatieRole($user)
 
             break;
         case 'project-developer':
+            $user->assignRole($user->role);
+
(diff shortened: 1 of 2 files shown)


app/functions.php
```diff
@@ -170,6 +170,9 @@ function assignSpatieRole($user)
 
             break;
         case 'project-developer':
+            $user->assignRole($user->role);
+
+            break;
```
roguenet, line 175: This change looks unnecessary - it's doing the same thing it was before, no?

roguenet (approved):

PR #324, opened 2024-07-02
[TM-951] feat: add stateMachine to site polygon

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-951)


app/Services/SiteService.php
```diff
@@ -0,0 +1,23 @@
...
+        if (! $site_uuid) {
+            return;
+        }
+        $site = Site::where('uuid', $site_uuid)->first();
+        if ($site->status != EntityStatusStateMachine::APPROVED) {
+            return;
+        }
+        $site->status = SiteStatusStateMachine::RESTORATION_IN_PROGRESS;
+        $site->save();
+    }
```
roguenet, line 22: This method is doing the work that the transitions and state machine methods should be doing instead. For one thing, the status should never be updated directly. Instead, it should be updated with `$site->status()->transitionTo()` (and then the `save()` call is not needed. 

The pattern established for the other state machines is to make this a method right on the model (or a trait for the model) as well. See for instance, `updateInProgress()` or `approve()` in `HasReportStatus`. Instead of having a service for this, I'd have the controller pull the record from the DB and call a method like `restorationInProgress()` on the `Site` instance, and that method would look something like
```
public function restorationInProgress()
{
  if ($site->status == EntityStatusStateMachine::RESTORATION_IN_PROGRESS) {
    $site->status()->transitionTo(SiteStatusStateMachine::RESTORATION_IN_PROGRESS);
  }
}
```

I don't love the check above for the current state. Ideally this method would only be called when the site _should_ be in the correct state, and then the `transitions` defined on the state machine will throw an error if the transition isn't allowed, but I don't know these flows well enough to know if that will cause problems.

app/StateMachines/SiteStatusStateMachine.php
```diff
@@ -0,0 +1,19 @@
...
+namespace App\StateMachines;
+
+class SiteStatusStateMachine extends EntityStatusStateMachine
+{
+    public const DRAFT = 'draft';
+    public const RESTORATION_IN_PROGRESS = 'restoration-in-progress';
+
+    public function transitions(): array
+    {
+        return [];
```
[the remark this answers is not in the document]
pachonjcl: changed

app/Models/Traits/HasSiteStatus.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+namespace App\Models\Traits;
+
+use App\Events\V2\General\EntityStatusChangeEvent;
+use App\StateMachines\SiteStatusStateMachine;
+use Asantibanez\LaravelEloquentStateMachines\Traits\HasStateMachines;
+use Illuminate\Support\Str;
+
+trait HasSiteStatus
```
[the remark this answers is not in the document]
pachonjcl: removed

app/Services/SiteService.php
pachonjcl: make use of `transitionTo`

app/StateMachines/SiteStatusStateMachine.php
```diff
@@ -0,0 +1,34 @@
...
+    public const DRAFT = 'draft';
+    public const AWAITING_APPROVAL = 'awaiting-approval';
+    public const APPROVED = 'approved';
+    public const NEEDS_MORE_INFORMATION = 'needs-more-information';
+    public const RESTORATION_IN_PROGRESS = 'restoration-in-progress';
+
+    public function recordHistory(): bool
+    {
+        return true;
+    }
```
roguenet, line 16: There's no reason to override this method with the same value already set in the super class.

app/StateMachines/SiteStatusStateMachine.php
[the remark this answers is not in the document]
pachonjcl: changed the transitions states to avoid using the parent ones.

app/StateMachines/SiteStatusStateMachine.php
```diff
@@ -0,0 +1,34 @@
...
+
+    public function transitions(): array
+    {
+        $parentTransitions = parent::transitions();
+
+        return array_merge(
+            $parentTransitions,
+            [
+                parent::APPROVED => [self::RESTORATION_IN_PROGRESS],
```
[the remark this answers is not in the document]
pachonjcl: Changed states
There is not difference really, just the name for UI usage.

app/StateMachines/SiteStatusStateMachine.php
pachonjcl: removed

app/StateMachines/SiteStatusStateMachine.php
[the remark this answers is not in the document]
roguenet: Ok, then please don't introduce a new state. The name mapping is specified in the `$statuses` member; you can simplly re-map the name there.

app/Services/SiteService.php
pachonjcl: name changed

app/StateMachines/SiteStatusStateMachine.php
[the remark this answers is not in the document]
LimberHope: As discussed yesterday we are keeping started

app/StateMachines/SiteStatusStateMachine.php
```diff
@@ -0,0 +1,24 @@
...
+            self::NEEDS_MORE_INFORMATION => [self::APPROVED, self::RESTORATION_IN_PROGRESS],
+            self::RESTORATION_IN_PROGRESS => [self::NEEDS_MORE_INFORMATION, self::APPROVED],
+            self::APPROVED => [self::NEEDS_MORE_INFORMATION, self::RESTORATION_IN_PROGRESS],
+        ];
+    }
+
+    public function defaultState(): ?string
+    {
+        return parent::STARTED;
+    }
```
roguenet, line 23: This override isn't needed because it's the same value as the parent.

app/StateMachines/SiteStatusStateMachine.php
```diff
@@ -0,0 +1,24 @@
...
+class SiteStatusStateMachine extends EntityStatusStateMachine
+{
+    public const RESTORATION_IN_PROGRESS = 'restoration-in-progress';
+
+    public function transitions(): array
+    {
+        return [
+            self::STARTED => [self::AWAITING_APPROVAL],
+            self::AWAITING_APPROVAL => [self::APPROVED, self::NEEDS_MORE_INFORMATION],
+            self::NEEDS_MORE_INFORMATION => [self::APPROVED, self::RESTORATION_IN_PROGRESS],
```
roguenet, line 14: This needs to continue to include `AWAITING_APPROVAL`. I think it could just be added here, but perhaps better way to accomplish what's being added here would be to supplement the super's transitions (example taken from `ReportStatusStateMachine`, which does something similar)
```
public function transitions(): array
{
    $parentTransitions = parent::transitions();
    // Reports can go from awaiting approval to started if the nothing_to_report flag is true (see validations below)
    $parentTransitions[self::NEEDS_MORE_INFORMATION][] = self::RESTORATION_IN_PROGRESS;

    return array_merge(
        [
            self::RESTORATION_IN_PROGRESS => [self::NEEDS_MORE_INFORMATION, self::APPROVED],
        ],
        $parentTransitions
    );
}
```

roguenet (approved):

PR #334, opened 2024-07-05
[TM-861] polygon versioning

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-861)


database/migrations/2024_07_05_140225_alter_site_polygon_table_to_add_versioning_attributes.php
```diff
@@ -0,0 +1,37 @@
...
+return new class extends Migration {
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('site_polygon', function (Blueprint $table) {
+            $table->string('version_name')->nullable();
+            $table->string('primary_uuid')->nullable()->after('uuid');
+            $table->integer('is_active')->nullable();
```
roguenet, line 16: This field would be better represented in the code as a boolean. I'd update this to `$table->boolean('is_active')->default(false)` and change the code that's using `1` and `0` for `is_active` to use `true` and `false`. It'll also result in a smaller column size, because it'll define the column as a `tinyint(1)`

database/migrations/2024_07_05_140225_alter_site_polygon_table_to_add_versioning_attributes.php
```diff
@@ -0,0 +1,37 @@
...
+            ;
+        });
+        DB::statement('UPDATE site_polygon set primary_uuid = uuid, is_active = 1;');
+
+        Schema::table('site_polygon', function (Blueprint $table) {
+            $table->uuid('primary_uuid')->nullable(false)->change();
```
roguenet, line 22: This column is used for lookups, and should get an index added.

app/Models/V2/User.php
```diff
@@ -321,4 +321,17 @@ public function wipeData()
...
+        if (empty($this->first_name) && empty($this->last_name)) {
+            return 'Unnamed User';
+        } elseif (empty($this->first_name)) {
+            return $this->last_name;
+        } elseif (empty($this->last_name)) {
+            return $this->first_name;
+        }
+
+        return $this->first_name . ' ' . $this->last_name;
+    }
```
roguenet, line 336: This attribute is fine, but is it used in this PR somewhere?

config/database.php
```diff
@@ -56,7 +56,7 @@
             'collation' => 'utf8mb4_unicode_ci',
             'prefix' => '',
             'prefix_indexes' => true,
-            'strict' => true,
+            'strict' => false,
```
roguenet, line 59: What does this do?

app/Models/V2/User.php
pachonjcl: Yes, I am using as `$user->full_name`

PR #345, opened 2024-07-10
[TM-1099] update site status

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-1099)

### DIFF
--- a/app/StateMachines/SiteStatusStateMachine.php
+++ b/app/StateMachines/SiteStatusStateMachine.php
@@ -12,6 +12,7 @@ class SiteStatusStateMachine extends EntityStatusStateMachine
 
         $parentTransitions[self::NEEDS_MORE_INFORMATION][] = self::RESTORATION_IN_PROGRESS;
         $parentTransitions[self::APPROVED][] = self::RESTORATION_IN_PROGRESS;
+        $parentTransitions[self::AWAITING_APPROVAL][] = self::RESTORATION_IN_PROGRESS;
 
         return array_merge(
             [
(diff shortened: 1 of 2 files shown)


app/StateMachines/SiteStatusStateMachine.php
```diff
@@ -12,6 +12,7 @@ public function transitions(): array
 
         $parentTransitions[self::NEEDS_MORE_INFORMATION][] = self::RESTORATION_IN_PROGRESS;
         $parentTransitions[self::APPROVED][] = self::RESTORATION_IN_PROGRESS;
+        $parentTransitions[self::AWAITING_APPROVAL][] = self::RESTORATION_IN_PROGRESS;
```
roguenet, line 15: If a site skips straight from awaiting approval to restoration in progress, the hooks in the state machine for `approved` won't go off. See `afterTransitionHooks` in the `EntityStatusStateMachine`. Perhaps what would be better would be to transition it to approved and then to restoration in progress to make sure the transition hooks go off as expected.

pachonjcl: image
My understanding is that restoration in progress is an intermediate state between awaiting approval and approved.

pachonjcl: So there is no need for firing off approved events.

roguenet: Oh, in that case is there any need to keep `approved` -> `restoration-in-progress`?

pachonjcl: It is exactly the rule that we need to address the new polygons creations that will immediately change a approved site to a "restoration-in-progress" site.

roguenet (approved):

PR #348, opened 2024-07-11
Feat/tm 973 making unit tests

egrojMonroy:
Add basic unit test for validators endpoints (geojson, shp, kml)


roguenet (approved):
👍🏻 Always happy to see the addition of unit tests.

PR #350, opened 2024-07-12
[TM-908] Project Manager role

roguenet:
https://gfw.atlassian.net/browse/TM-908
[FE PR](https://github.com/wri/wri-terramatch-website/pull/376)


app/Models/V2/Projects/Project.php
```diff
@@ -312,6 +312,11 @@ public function users(): BelongsToMany
         return $this->belongsToMany(User::class, 'v2_project_users')->withPivot(['status', 'is_monitoring']);
     }
 
+    public function managers(): BelongsToMany
+    {
+        return $this->belongsTomany(User::class, 'v2_project_users')->wherePivot('is_managing', true);
```
pachonjcl, line 317: is this a typo or this is intended? `belongsTomany`

roguenet: Oh wild, today I learned that PHP function names are case insensitive. What an "interesting" language 😂 . Fixing now.

pachonjcl (approved):

PR #351, opened 2024-07-15
[TM-840] add condition when projectIds < 0

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-840)

### DIFF
--- a/app/Http/Controllers/V2/User/IndexMyActionsController.php
+++ b/app/Http/Controllers/V2/User/IndexMyActionsController.php
@@ -15,16 +15,12 @@ class IndexMyActionsController extends Controller
     {
         $user = Auth::user();
 
-        $organisationId = $user->organisation_id;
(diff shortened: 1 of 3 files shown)


roguenet (dismissed):

app/Http/Controllers/V2/User/IndexMyActionsController.php
```diff
@@ -22,15 +22,15 @@ public function __invoke(Request $request): AnonymousResourceCollection
             ->with('targetable')
             ->pending();
 
+        $qry->where('project_id', $projectIds);
```
roguenet, line 25: I'm curious about this change; wasn't the previous approach more correct? If there are multiple project ids, this call would need to be `whereIn` as on line 29.

roguenet (dismissed):

roguenet (approved):

PR #352, opened 2024-07-16
[TM-861] add polygon version

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-861)

### DIFF
--- a/app/Http/Controllers/V2/Sites/UpdateSitePolygonActiveController.php
+++ b/app/Http/Controllers/V2/Sites/UpdateSitePolygonActiveController.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace App\Http\Controllers\V2\Sites;
+
+use App\Http\Controllers\Controller;
+use App\Models\V2\Sites\SitePolygon;
+use Illuminate\Http\JsonResponse;
+use Illuminate\Support\Facades\Log;
+
+class UpdateSitePolygonActiveController extends Controller
+{
+    public function __invoke(string $uuid): JsonResponse
+    {
+        Log::info('Fetching site polygons', ['uuid' => $uuid]);
+
+        try {
+            $sitePolygon = SitePolygon::isUuid($uuid)->first();
+            if (! $sitePolygon) {
+                return response()->json(['error' => 'Site polygon not found'], 404);
+            }
+            SitePolygon::where('primary_uuid', $sitePolygon->primary_uuid)->update(['is_active' => 0]);
+
+            $sitePolygon->is_active = true;
+            $sitePolygon->save();
+
+            return response()->json($sitePolygon);
+        } catch (\Exception $e) {
+            Log::error($e->getMessage());
+
+            return response()->json(['error' => 'An error occurred while fetching site polygons'], 500);
+        }
+    }
+}
--- a/app/Http/Controllers/V2/Sites/UpdateSitePolygonActiveController.php
+++ b/app/Http/Controllers/V2/Sites/UpdateSitePolygonActiveController.php
@@ -18,7 +18,7 @@ class UpdateSitePolygonActiveController extends Controller
             if (! $sitePolygon) {
                 return response()->json(['error' => 'Site polygon not found'], 404);
             }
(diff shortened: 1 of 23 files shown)


database/migrations/2024_07_05_140225_alter_site_polygon_table_to_add_versioning_attributes.php
```diff
@@ -0,0 +1,38 @@
...
+return new class extends Migration {
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('site_polygon', function (Blueprint $table) {
+            $table->string('version_name')->nullable();
+            $table->string('primary_uuid')->nullable()->after('uuid');
+            $table->boolean('is_active')->default(false);
```
pachonjcl, line 16: This one is not a boolean and references has been updated to true / false accordingly.

database/migrations/2024_07_05_140225_alter_site_polygon_table_to_add_versioning_attributes.php
```diff
@@ -0,0 +1,38 @@
...
+        });
+
+        DB::statement('UPDATE site_polygon set primary_uuid = uuid, is_active = true;');
+
+        Schema::table('site_polygon', function (Blueprint $table) {
+            $table->uuid('primary_uuid')->nullable(false)->change();
+            $table->index('primary_uuid');
```
pachonjcl, line 23: There is not an index on primary_uuid

config/database.php
```diff
@@ -56,7 +56,7 @@
             'collation' => 'utf8mb4_unicode_ci',
             'prefix' => '',
             'prefix_indexes' => true,
-            'strict' => true,
+            'strict' => false,
```
pachonjcl, line 59: Making reference of this [SO post](https://stackoverflow.com/questions/47974348/sqlstatehy000-general-error-1364-field-name-doesnt-have-a-default-value-l)
Unit tests were throwing errors as follows:
`PDOException: SQLSTATE[HY000]: General error: 1364 Field 'primary_uuid' doesn't have a default value`
This is the difference between strict usage and not.

app/Http/Controllers/V2/Sites/UpdateSitePolygonActiveController.php
```diff
@@ -0,0 +1,33 @@
...
+    public function __invoke(string $uuid): JsonResponse
+    {
+        Log::info('Fetching site polygons', ['uuid' => $uuid]);
+
+        try {
+            $sitePolygon = SitePolygon::isUuid($uuid)->first();
+            if (! $sitePolygon) {
+                return response()->json(['error' => 'Site polygon not found'], 404);
+            }
+            SitePolygon::where('primary_uuid', $sitePolygon->primary_uuid)->update(['is_active' => 0]);
```
roguenet, line 21: This should use `=> false`

app/Models/V2/Sites/SitePolygon.php
```diff
@@ -97,4 +101,39 @@ public function getAuditableNameAttribute(): string
     {
         return $this->poly_name ?? '';
     }
+
+    public function scopeActive(Builder $query)
+    {
+        return $query->where('is_active', 1);
```
roguenet, line 107: `where('is_active', true)`

database/migrations/2024_07_05_140225_alter_site_polygon_table_to_add_versioning_attributes.php
roguenet, line 16: On reading this again, I think maybe the default should be true?

app/Models/V2/Sites/SitePolygon.php
pachonjcl: changed

app/Http/Controllers/V2/Sites/UpdateSitePolygonActiveController.php
pachonjcl: changed

roguenet (approved):

PR #353, opened 2024-07-17
[TM-1087] Polygon Version (PD View Site Profile)

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-1087)


roguenet (commented):
This code change seems fine, but I'm concerned about the huge number of commits that look like they're merging in. That tells me that something weird happened with the git history on this branch. I think maybe re-create the branch from staging and just bring in this change to make sure this doesn't junk up the history on `staging`.

PR #356, opened 2024-07-19
[TM-1079] transform polygon applications

egrojMonroy:
https://gfw.atlassian.net/browse/TM-1079

This commint includes two commands:
php artisan parse:pitch-boundaries {framework_key}
php artisan parse:project-boundaries {framework_key}

Both parse the geojson attribute inside each table, and insert them into project_polygon table with the geojson parsed through convexhull algorithm to obtain one single polygon. 
Some geojson that has one or two points, have to be parsed through voronoi algorithm, all points have est_area = 78. And then applied convexhull.


app/Console/Commands/ParseGeomBoundariesPitches.php
```diff
@@ -0,0 +1,78 @@
...
+    protected $description = 'Parse string geojsons of projects and project pitches by framework_key';
+
+    public function handle()
+    {
+        $frameworkKey = $this->argument('framework_key');
+
+        $projectPitches = $this->getProjectPitches($frameworkKey);
+        $ids = $projectPitches->pluck('id')->toArray();
+        $this->info(json_encode($ids));
+        $bar = $this->output->createProgressBar(count($projectPitches));
```
roguenet, line 27: Oh this is super handy, I didn't know `Command` had this capability.

app/Console/Commands/ParseGeomBoundariesPitches.php
```diff
@@ -0,0 +1,78 @@
...
+        $this->info("\nGeometry boundaries parsing completed.");
+    }
+
+    private function getProjectPitches($frameworkKey)
+    {
+        return ProjectPitch::whereHas('fundingProgramme', function ($query) use ($frameworkKey) {
+            $query->where('framework_key', $frameworkKey);
+        })
+        ->whereNotNull('proj_boundary')
+        ->where('proj_boundary', '!=', 'null')
```
roguenet, line 47: It's so obnoxious that we have the literal string value "null" in this column, but I checked and there are 369 in prod 🤷🏻

roguenet (dismissed):

cesarLima1 (approved):

PR #359, opened 2024-07-23
[TM-1139, 1140] add verify invites to monitoring partners

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-1139)
[Task Link](https://gfw.atlassian.net/browse/TM-1140)

### DIFF
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -42,7 +42,7 @@ class AuthController extends Controller
         }
         $invites = ProjectInvite::where('email_address', $me->email_address)->get();
         foreach ($invites as $invite) {
-            $me->projects()->sync([$invite->project_id => ['is_monitoring' => true]]);
+            $me->projects()->sync([$invite->project_id => ['is_monitoring' => true]], false);
             if ($me->organisation_id == null) {
                 $me->organisation_id = $invite->project->organisation_id;
                 $me->saveOrFail();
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/AuthController.php
```diff
@@ -40,6 +40,18 @@ public function loginAction(LoginRequest $request): JsonResponse
         if (is_null($me->password)) {
             throw new FailedLoginException();
         }
+        $invites = ProjectInvite::where('email_address', $me->email_address)->get();
+        foreach ($invites as $invite) {
+            $me->projects()->sync([$invite->project_id => ['is_monitoring' => true]]);
```
roguenet, line 45: I believe you should be passing `false` in here as a second parameter to prevent this call from removing the other associated projects (including those added in previous iterations of this loop).

app/Http/Controllers/AuthController.php
```diff
@@ -40,6 +40,18 @@ public function loginAction(LoginRequest $request): JsonResponse
...
+            if ($me->organisation_id == null) {
+                $me->organisation_id = $invite->project->organisation_id;
```
roguenet, line 47: Should this really be setting the the org to the last one from the list of invites? This code feels like something that isn't supposed to be in a loop; are we expecting there to be multiple invites for a user? If not, perhaps this should throw an error if there are multiple instead of what it's doing here.

LimberHope: A similar situation occurred once there was more than one invitation for a user and that is why this logic was implemented, in fact it is similar to the "verifyUnauthorizedAction" function found in "AuthController.php" but this time it is being used at login.

roguenet (approved):

PR #366, opened 2024-07-24
[TM-1106] Improved long term scheduling system.

roguenet:
https://gfw.atlassian.net/browse/TM-1106


config/cache.php
```diff
@@ -15,7 +15,7 @@
     |
     */
 
-    'default' => env('CACHE_DRIVER', 'file'),
+    'default' => env('CACHE_DRIVER', 'redis'),
```
roguenet, line 18: Switching to the redis cache is necessary for using the `->onOneServer()` functionality in Kernel.php. One of the big problems with scheduling before was that we'd often times end up with quadruplicate generated reports, and that's because the task was running on all four of our production nodes (two web, two worker). This should prevent that happening any longer, although it'll need testing on staging. We were also getting quadruplicate reports generated, which doesn't have a ton of consequences, but does soak up a bunch of extra resources for no good reason.

app/Models/V2/ScheduledJobs/TaskDueJob.php
```diff
@@ -0,0 +1,130 @@
...
+                'due_at' => $dueAt,
+            ],
+        ]);
+    }
+
+    public function scopeFramework(Builder $query, string $frameworkKey): Builder
+    {
+        return $query->whereJsonContains('task_definition->framework_key', $frameworkKey)
+            ->orderBy('execution_time');
+    }
```
roguenet, line 39: This scope makes it easy to determine the next time a given framework is set to either generate reports, or when those reports will be due (in conjunction with the `getDueAtAttribute` below):
```
> TaskDueJob::framework('terrafund')->first()->execution_time
= Illuminate\Support\Carbon @1735603200 {#9010
    date: 2024-12-31 00:00:00.0 UTC (+00:00),
  }

> TaskDueJob::framework('terrafund')->first()->due_at
= Carbon\Carbon @1738281600 {#9030
    date: 2025-01-31 00:00:00.0 +00:00,
  }
```

app/Console/Commands/OneOff/SeedScheduledJobs.php
```diff
@@ -0,0 +1,84 @@
...
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Creates the initial definition of scheduled jobs to replace what was in Kernel.php';
+
+    /**
+     * Execute the console command.
+     */
+    public function handle()
```
roguenet, line 30: This task duplicates the scheduled jobs removed from `Kernel.php`. The assumption is that by the time we've generated the reports seeded here, we'll have the new UI in place to set new execution and report generation times and won't need to do it by Command, or manually in tinker any longer.

cesarLima1 (approved):

cesarLima1:
LGTM. 
Testing will take time if we want to test the scheduling. It will be interesting to see if reports are generated without duplicates and if there are any reports that were not generated.

Scriptmatico (approved):

PR #379, opened 2024-07-31
[TM-1144] reorder functions to get valid schema data for polygons

cesarLima1:
https://gfw.atlassian.net/browse/TM-1144

### DIFF
--- a/app/Services/PolygonService.php
+++ b/app/Services/PolygonService.php
@@ -14,12 +14,12 @@ use App\Models\V2\Sites\SitePolygon;
 use App\Models\V2\User;
 use App\Validators\SitePolygonValidator;
 use DateTime;
+use Exception;
 use Illuminate\Support\Carbon;
 use Illuminate\Support\Facades\App;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\DB;
 use InvalidArgumentException;
-use Exception;
 
 class PolygonService
 {
@@ -389,6 +389,7 @@ class PolygonService
     {
         try {
             $d = DateTime::createFromFormat('Y-m-d', $date);
+
             return $d && $d->format('Y-m-d') === $date;
         } catch (Exception $e) {
             return false;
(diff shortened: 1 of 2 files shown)


app/Services/PolygonService.php
```diff
@@ -323,4 +347,59 @@ protected function transformAndStorePoints($geojson, $sitePolygonProperties): ar
...
+
+    public function isFieldInvalid($field, $value)
+    {
+        if (is_null($value) || $value === '') {
+            return true;
+        }
+
+        switch ($field) {
+            case 'plantstart':
+                return ! $this->isValidDate($value);
```
pachonjcl, line 359: If you take a close look, I don't think you are following the same method naming.
you have `isSomethingValid` vs `isValidSomething`. 
My suggestion is to have `isInvalidField`, but do not mix

app/Services/PolygonService.php
```diff
@@ -323,4 +347,59 @@ protected function transformAndStorePoints($geojson, $sitePolygonProperties): ar
...
+
+        return true;
+    }
+
+    private function isValidDate($date)
+    {
+        $d = DateTime::createFromFormat('Y-m-d', $date);
+
+        return $d && $d->format('Y-m-d') === $date;
+    }
```
pachonjcl, line 404: isn't it possible that the DateTime parse throws an error? can you catch in the flow?

app/Services/PolygonService.php
```diff
@@ -323,4 +347,59 @@ protected function transformAndStorePoints($geojson, $sitePolygonProperties): ar
...
+    {
+        $distributions = explode(',', $value);
+        foreach ($distributions as $distribution) {
+            if (! in_array(trim($distribution), self::VALID_DISTRIBUTIONS)) {
+                return false;
+            }
+        }
+
+        return true;
+    }
```
pachonjcl, line 397: nitpick, but this two methods are exactly the same with two different parameters.

pachonjcl (changes requested):

app/Services/PolygonService.php
cesarLima1: Done.

app/Services/PolygonService.php
cesarLima1: try catch added.

app/Services/PolygonService.php
cesarLima1: Using the same function for both now.

pachonjcl (approved):

PR #390, opened 2024-08-05
[TM-1179] revert approved on update geom

egrojMonroy:
https://gfw.atlassian.net/jira/software/c/projects/TM/boards/40?selectedIssue=TM-1179

### DIFF
--- a/app/Helpers/GeometryHelper.php
+++ b/app/Helpers/GeometryHelper.php
@@ -307,14 +307,6 @@ class GeometryHelper
         return $result[0]->wkt ?? null;
     }
 
-    public static function updateStatusOnEdit($sitePolygon)
-    {
-        if ($sitePolygon->status === 'approved') {
-            $sitePolygon->status = 'submitted';
-            $sitePolygon->save();
-        }
-    }
-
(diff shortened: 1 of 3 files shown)


app/Helpers/GeometryHelper.php
```diff
@@ -307,6 +307,14 @@ public static function getConvexHull($geoJson)
         return $result[0]->wkt ?? null;
     }
 
+    public static function updateStatusOnEdit($sitePolygon)
+    {
+        if ($sitePolygon->status === 'approved') {
+            $sitePolygon->status = 'submitted';
+            $sitePolygon->save();
+        }
+    }
```
pachonjcl, line 316: I think this function would be better in `app/Models/V2/Sites/SitePolygon.php` and then name it as `edit` or something like this.
`$sitePolygon->edit()`

pachonjcl (changes requested):

app/Helpers/GeometryHelper.php
egrojMonroy: right, moved to model.

pachonjcl (approved):

PR #392, opened 2024-08-07
[TM-578] store user locale preference

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-578)

This PR creates a new Table called `localization_keys`, with the following attributes:

- id
- key
- value
- value_id
- tag

Each LocalizationKey Has a 1 - 1 relationship with I18nItem through value_id.

The purpose of the LocalizationKey is to store a value that needs translation and each time a new LocalizationKey is added should a I18nItem should be created.

Right now, all the localization keys needed for translating emails are stored in `database/seeders/AddLocalizationKeys.php` file which is a seeder.

The usage of this `LocalizationKeys` is made by `I18nMail` which is just a wrapper for Mail that has setters for the following properties:

- subjectKey
- titleKey
- bodyKey
- ctaKey

Along
- userLocale
And any param that will be used in the translation (like project names or counts)
- params

`I18nMail` will take those key parameters and translate based on the `LocalizationKey` and pass to `Mail` itself and everything else should keep the same.


app/Console/Commands/SendTerrafundReportRemindersCommand.php
```diff
@@ -32,10 +32,13 @@ public function handle(): int
             ->chunkById(100, function ($programmes) {
                 $programmes->each(function ($programme) {
                     if ($programme->users->count()) {
-                        Mail::to($programme->users->pluck('email_address'))->queue(new TerrafundReportReminder($programme->id));
                         $programme->users->each(function ($user) use ($programme) {
+                            Mail::to($user->email_address)
```
roguenet, line 36: 🤔 I understand why you had to do this, but it's going to increase our email send volume quite a bit to break apart emails that used to go to several people into individual emails. We might need to look into rate limiting our email send jobs, or handling errors from SES and re-trying the send later, if that's not already wired up. Nothing to do about it in this PR, but something for us to keep an eye on in Sentry.

routes/api_v2.php
```diff
@@ -741,3 +742,5 @@ function () {
 });
 
 Route::get('/type-entity', EntityTypeController::class);
+
+Route::post('/locale/{locale}', UserLocaleController::class);
```
roguenet, line 746: I already approved the FE PR, but I think this needs a change. A REST structure like this would indicate to me that the FE is trying to create a Locale, which is weird. Also, for a Post, I would expect the value to be in the body, not the URL. We're going to be doing a lot of work to make our API actually RESTful in the v3 architecture, but the fewer missteps we have in new APIs in v2, the easier that will be.

I think this should instead be structured as something more like 
```
PATCH /api/v2/user
```
or 
``` 
PATCH /api/v2/user/locale
```
with the locale value in the body. If we already had a proper user update endpoint, I'd say this should just be part of that patch, but for instance password reset is done through `PATCH /api/v2/user/change`, which is also very weird. Given that, I think `PATCH /api/v2/user/locale` is (while still a bit strange), the less confusing option above given the current state of the PHP API.

Ideally (in v3), we'll have a simple patch user endpoint (likely something like `PATCH /user-service/user/me` with `me` being a "magic" ID so we don't have to pass around the current user's ID in the API) that can take payloads with various things that can be updated (like password and locale) and then returns the full updated Me user resource as the response.

app/Http/Controllers/V2/UserLocaleController.php
```diff
@@ -0,0 +1,22 @@
...
+namespace App\Http\Controllers\V2;
+
+use App\Http\Controllers\Controller;
+use App\Http\Requests\V2\UserLocaleRequest;
+use Illuminate\Http\JsonResponse;
+use Illuminate\Support\Facades\Auth;
+
+class UserLocaleController extends Controller
+{
+    public function __invoke(UserLocaleRequest $request, string $locale): JsonResponse
```
roguenet, line 12: Why is the locale coming both from the request and as a separate parameter on the controller? If we were to continue keeping it in the URL, I think `string $locale` should be dropped here, and it should be pulled from the request instead. More notes about this structure in my comment in `api_v2.php`

roguenet (commented):
Good work on building the translation system for the emails 👍🏻 . My only comments are around the API structure for setting the user's locale.

app/Http/Controllers/V2/UserLocaleController.php
pachonjcl: Right now if you review the `UserLocaleRequest` we are not using the `locale` from the body, we are using the one on the route param.
To validate that it is one of the languages we have available.

app/Console/Commands/SendTerrafundReportRemindersCommand.php
pachonjcl: sure, will keep in mind

app/Http/Controllers/V2/UserLocaleController.php
pachonjcl: I now understand what you mentioned, I just obtained from the request object.

LimberHope (approved):

PR #398, opened 2024-08-10
[TM-1103] Update user roles and permissions

roguenet:
https://gfw.atlassian.net/browse/TM-1103
[FE PR](https://github.com/wri/wri-terramatch-website/pull/429)

The main purpose of this work is to remove the `users.role` column, and rely solely on the Spatie roles system for permissions. The v1 `User` and `Admin` models have also been removed, and a few controllers that are no longer in use have been removed as well.

For adding roles and permissions in the future, please modify `config/wri/permissions.php`, and then run `php artisan sync-roles`. 

In order to get this working locally, make sure to run the `php artisan one-off:move-user-roles-to-v2` script.


cesarLima1:
LGTM. 
This PR should probably be flagged in the Slack channel, as the documentation might need updating after removing `role` from `user` table.

cesarLima1 (dismissed):

cesarLima1 (approved):

PR #411, opened 2024-08-16
Feat/[TM 1185] clip polygons 

egrojMonroy:
https://gfw.atlassian.net/browse/TM-1185


app/Helpers/PolygonGeometryHelper.php
```diff
@@ -58,4 +58,13 @@ public static function updateProjectCentroidFromPolygon($polygonGeometry)
             Log::error('Error updating project centroid: ' . $e->getMessage());
         }
     }
+
+    public static function getPolygonsWithNames(array $uuids)
```
pachonjcl, line 62: I don't think this method is going to be reused anytime soon the way it is, my suggestion would be to call this as `getPolygonsProjection` and pass parameters `$fields` that will be the current array of fields that you want to project your site polygon

pachonjcl: Also I noted that your method does not take the `is_active` scope or filter into account. Please confirm that this is the intended behavior

app/Http/Controllers/V2/Terrafund/TerrafundClipGeometryController.php
```diff
@@ -0,0 +1,72 @@
...
+    {
+        $polygonUuids = GeometryHelper::getSitePolygonsUuids($uuid)->toArray();
+
+        return $this->processClippedPolygons($polygonUuids);
+    }
+
+    public function clipOverlappingPolygons(string $uuid)
+    {
+        $polygonOverlappingExtraInfo = CriteriaSite::where('polygon_id', $uuid)
+            ->where('criteria_id', PolygonService::OVERLAPPING_CRITERIA_ID)
```
pachonjcl, line 27: create a scope in the CriteriaSite Model to only pass a CriteriaID as a parameter

pachonjcl (changes requested):

app/Helpers/PolygonGeometryHelper.php
egrojMonroy: Changed to be more usefull

egrojMonroy: also added is_active

app/Http/Controllers/V2/Terrafund/TerrafundClipGeometryController.php
egrojMonroy: using a already created scope now

pachonjcl (approved):

PR #423, opened 2024-08-21
[TM-1171] fix role policy check

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-1171)


roguenet (commented):
Please use `can('projects-manage')` over `hasRole`. We should always prefer to check permissions over roles in the updated system, when possible.

roguenet (approved):

roguenet (changes requested):
Whoops, meant to request changes. Please use `can->('projects-manage')` instead.

PR #463, opened 2024-09-12
[TM-1253] Search refactor for TM-1253, TM-1254, TM-1255, TM-1256, TM-1257

Scriptmatico:
https://gfw.atlassian.net/browse/TM-1246

### DIFF
--- a/app/Models/V2/Forms/Application.php
+++ b/app/Models/V2/Forms/Application.php
@@ -65,9 +65,10 @@ class Application extends Model
         ];
     }
 
-    public static function searchApplications($query){
+    public static function searchApplications($query)
+    {
         return self::select('applications.*')
-            ->join('organisations', 'applications.organisation_uuid','=','organisations.uuid')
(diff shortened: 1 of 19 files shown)


app/Models/V2/Forms/Application.php
```diff
@@ -13,14 +13,12 @@
...
 use Illuminate\Database\Eloquent\Relations\HasOne;
 use Illuminate\Database\Eloquent\SoftDeletes;
-use Laravel\Scout\Searchable;
 
 class Application extends Model
 {
     use HasFactory;
     use HasUuid;
     use SoftDeletes;
-    use Searchable;
```
roguenet, line 23: 🎨 This approach is fine and you can leave as-is if you prefer it this way, but I'd probably have made the new method with the name `search` for brevity (Application::searchApplications is a little duplicative). With this trait being removed, there shouldn't be any problem with that name.

roguenet (commented):
I'm ready to approve once the checks on the PR pass.

roguenet (approved):

PR #466, opened 2024-09-16
[TM-1278] Add task daily digest

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-1278)


app/Console/Commands/SendDailyDigestNotifications.php
```diff
@@ -0,0 +1,34 @@
...
+     * @var string
+     */
+    protected $description = 'Runs daily digest notifications for tasks';
+
+    /**
+     * Execute the console command.
+     */
+    public function handle()
+    {
+        Task::isIncomplete()->get()->each(function (Task $task) {
```
roguenet, line 30: Tasks are pretty small, but there are 363 in `Task::isIncomplete()` in my current copy of prod. Better to do this in chunks instead of loading them all into memory at once. `chunkById` is what you want, and there are several examples of its use in the codebase, especially in the Commands code.

app/Jobs/SendDailyDigestNotificationsJob.php
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Execute the job.
+     */
+    public function handle(): void
+    {
+        $users = $this->task->project->users()->get();
+        foreach ($users as $user) {
+            Mail::to($user->email_address)->send(
+                new TaskDigestMail($user, $this->task)
```
roguenet, line 39: This is going to result in a huge volume of email. I would be shocked if we don't get API errors from Amazon and likely need to space these sends out in delayed chunks, which is a pretty big change for this PR. Just dropped a question in Slack to see if we can delay this feature till the next release.

resources/views/Emails/master.blade.php
```diff
@@ -2,6 +2,18 @@
...
         <meta charset="utf-8">
+        <!-- add styles -->
+        <style>
+            .table-custom {
+                table-layout : fixed;
+                border-collapse: collapse;
+                border-spacing: 0;
+                width: 100%;
+            }
+            .border-custom {
```
roguenet, line 13: These would do better with a more descriptive name than `-custom`. What's custom about them? What do they look like? Or what purpose are they for?

roguenet (commented):
I was expecting this PR to also remove the current transactional emails that get sent today, not just add the digest. I'm pretty sure that's meant to be part of this ticket as well.

roguenet:
Looks like there are conflicts to resolve if we do end up including this in the release.

app/Jobs/SendDailyDigestNotificationsJob.php
```diff
@@ -34,10 +34,11 @@ public function __construct(Task $task)
     public function handle(): void
     {
         $users = $this->task->project->users()->get();
-        foreach ($users as $user) {
-            Mail::to($user->email_address)->send(
-                new TaskDigestMail($user, $this->task)
-            );
+        $usersGroupedByLocale = $users->groupBy('locale');
+
+        foreach ($usersGroupedByLocale as $locale => $users) {
```
roguenet, line 39: Excellent improvement 👍🏻

roguenet (approved):

pachonjcl:
According to the conversation in the ticket, the frequency is going to change and probably some other logic needs to be defined. This should not be merged for now.

LimberHope (approved):

PR #500, opened 2024-10-09
[TM-1127] Add worker

egrojMonroy:
https://gfw.atlassian.net/browse/TM-1127

Here are the workers for three proceesses:

- Polygon upload
- Bulk polygon validation
- Fix polygons 

Also the [task ](https://gfw.atlassian.net/browse/TM-1327) is merged in this branch, because it is needed to make the fix polygons Job. 

This is mark as draft because of the dependent [ticket](https://gfw.atlassian.net/browse/TM-1354) 

[v3 PR](https://github.com/wri/terramatch-microservices/pull/7)
[FE PR](https://github.com/wri/wri-terramatch-website/pull/596)


app/Http/Controllers/V2/Terrafund/TerrafundClipGeometryController.php
```diff
@@ -2,35 +2,53 @@
...
 {
     public function clipOverlappingPolygonsBySite(string $uuid)
     {
+        $user = Auth::user();
         $polygonUuids = GeometryHelper::getSitePolygonsUuids($uuid)->toArray();
+        $job = new FixPolygonOverlapJob($polygonUuids, $user->id);
+        $jobUUID = $job->getJobUuid();
+        dispatch($job);
+
+        return response()->json(['job_uuid' => $jobUUID], 200);
```
roguenet, line 26: Let's enforce consistency on these with a Resource class. Something like `DelayedJobResource` that gets returned from each of these endpoints.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -418,7 +377,7 @@ public function uploadShapefile(Request $request)
                 $polygonLoadedList = isset($body['polygon_loaded']) && filter_var($body['polygon_loaded'], FILTER_VALIDATE_BOOLEAN);
                 $submitPolygonsLoaded = $request->input('submit_polygon_loaded') === true || $request->input('submit_polygon_loaded') === 'true';
                 if (! $polygonLoadedList && ! $submitPolygonsLoaded) {
-                    $uuid = $this->insertGeojsonToDB($geojsonFilename, $site_id, 'site', $body['primary_uuid'] ?? null);
+                    $job = new InsertGeojsonToDBJob($geojsonFilename, $site_id, 'site', $body['primary_uuid'] ?? null);
```
roguenet, line 380: Remember that in AWS, the workers and the web boxes are separate EC2 machines and do not share a file system. This geoJson file is going to need to get uploaded to an S3 bucket and then pulled down in the job.

app/Jobs/FixPolygonOverlapJob.php
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     * Execute the job.
+     */
+    public function handle(): void
+    {
+
+        try {
+            DelayedJob::create([
+              'uuid' => $this->job_uuid,
```
roguenet, line 72: Creating the UUID should be handled by the HasUuid trait, which needs to be added to DelayedJob. There should be no need to create them manually.

Furthermore, the Delayed job should be created when this Job is created, not when it gets started on the worker box in `handle`. The API should be able to find a row in the DB for this job as soon as the UUID is returned from the original controller; the way this is set up now, it's creating a race condition that could cause a 404 to get returned from the v3 delayed jobs endpoint.

app/Jobs/FixPolygonOverlapJob.php
```diff
@@ -0,0 +1,110 @@
...
+              'status' => self::STATUS_PENDING,
+              'created_at' => now(),
+            ]);
+            $user = Auth::user();
+            if ($user) {
+                $polygonsClipped = App::make(PolygonService::class)->processClippedPolygons($this->polygonUuids);
+                DelayedJob::where('uuid', $this->job_uuid)->update([
+                  'status' => self::STATUS_SUCCEEDED,
+                  'payload' => json_encode(['updated_polygons' => $polygonsClipped]),
+                  'updated_at' => now(),
```
roguenet, line 82: We should never be managing these timestamps ourselves. The HasTimestamps trait that's included in the base Model class does this for us.

app/Models/DelayedJob.php
```diff
@@ -0,0 +1,58 @@
...
+     * Set the status of the job.
+     *
+     * @param string $status
+     * @return void
+     */
+    public function setStatus(string $status): void
+    {
+        $this->status = $status;
+        $this->save();
+    }
```
roguenet, line 45: Are these used? This is a non-idiomatic way to handle attributes in laravel. The code that has a reference to this class should just call `$delayedJob->status` directly, and `$delayedJob->update(['status' => $status])`

roguenet: Same question for `updateStatusCode` below; I think these three methods are superfluous.

database/migrations/2024_10_03_150604_create_delayed_jobs_table.php
```diff
@@ -0,0 +1,30 @@
...
+return new class extends Migration {
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('delayed_jobs', function (Blueprint $table) {
+            $table->id();
+            $table->uuid('uuid')->unique();
+            $table->enum('status', ['pending', 'failed', 'succeeded']);
```
roguenet, line 16: We do use them in a few places, but `enum` isn't really recommended for typical DB usage because it's hard to modify or remove alues without data loss. This can be a simple string.

database/migrations/2024_10_03_150604_create_delayed_jobs_table.php
```diff
@@ -0,0 +1,30 @@
...
+            $table->integer('statusCode')->nullable();
+            $table->string('payload');
```
roguenet, line 18: This should be a `json` and needs to be nullable.

roguenet (commented):
In addition to the comments so far, this needs API documentation updates. All these endpoints that were modified need to have a simple delayed job response shape. We'll want to codify the possible types of delayed job responses in the V3 endpoint documentation.

database/migrations/2024_10_03_150604_create_delayed_jobs_table.php
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('delayed_jobs', function (Blueprint $table) {
+            $table->id();
+            $table->uuid('uuid')->unique();
+            $table->enum('status', ['pending', 'failed', 'succeeded']);
+            $table->integer('statusCode')->nullable();
```
roguenet, line 17: This should be camel_case, as it's the standard for all of our DB table columns.

egrojMonroy: changed

database/migrations/2024_10_03_150604_create_delayed_jobs_table.php
egrojMonroy: done

database/migrations/2024_10_03_150604_create_delayed_jobs_table.php
egrojMonroy: done

app/Models/DelayedJob.php
egrojMonroy: removed

app/Jobs/FixPolygonOverlapJob.php
egrojMonroy: done

app/Http/Controllers/V2/Terrafund/TerrafundClipGeometryController.php
egrojMonroy: done

app/Jobs/FixPolygonOverlapJob.php
roguenet: Hmm, it looks like the `updated_at' => now()` line is still there in this instance.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -220,28 +222,31 @@ function ($attribute, $value, $fail) {
...
-            $polygonLoaded = $this->GetAllPolygonsLoaded($geojsonContent, $site_id);
-
-            return response()->json($polygonLoaded->original, 200);
+          $polygonLoaded = $this->GetAllPolygonsLoaded($geojsonContent, $site_id);
+          return response()->json($polygonLoaded->original, 200);
         }
 
         if ($submitPolygonsLoaded) {
-            $job = new InsertGeojsonToDBJob($filename, $site_id, 'site', $body['primary_uuid'] ?? null, $body['submit_polygon_loaded']);
+          $job = new InsertGeojsonToDBJob($redis_key, $delayedJob->id, $site_id, 'site', $body['primary_uuid'] ?? null, $body['submit_polygon_loaded']);
```
roguenet, line 244: It took me a second to decipher these three if blocks because at first glance it looks like a job can be created and then not used in the return response. I'd restructure a bit to make it more readable. Also, it looks like the job created on 235 and here are the same, and that a DelayedJob can end up getting created (and a Redis payload set) that doesn't get used.

```
if ($polygonLoadedList) {
  // ... return loaded response
} 

// insert payload to Redis, create DelayedJob and InsertGeojsonToDBJob
```

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -373,28 +377,29 @@ public function uploadShapefile(Request $request)
...
-                    $filePath = $tempDir . DIRECTORY_SEPARATOR . $geojsonFilename;
-                    $geojsonContent = file_get_contents($filePath);
                     $polygonLoaded = $this->GetAllPolygonsLoaded($geojsonContent, $site_id);
-
                     return response()->json($polygonLoaded->original, 200);
                 }
 
                 if ($submitPolygonsLoaded) {
-                    $job = new InsertGeojsonToDBJob($geojsonFilename, $site_id, 'site', $body['primary_uuid'] ?? null, $body['submit_polygon_loaded']);
+                    $job = new InsertGeojsonToDBJob($redis_key, $delayedJob->id,  $site_id, 'site', $body['primary_uuid'] ?? null, $body['submit_polygon_loaded']);
```
roguenet, line 398: Looks like this method has the same structure problems as the one above.

app/Jobs/FixPolygonOverlapJob.php
egrojMonroy: Job Created before running te worker now.

app/Jobs/InsertGeojsonToDBJob.php
```diff
@@ -15,94 +15,76 @@
...
     public $timeout = 0;
 
-    private const STATUS_PENDING = 'pending';
-    private const STATUS_FAILED = 'failed';
-    private const STATUS_SUCCEEDED = 'succeeded';
-
-    protected $geojsonFilename;
+    public const STATUS_PENDING = 'pending';
+    public const STATUS_FAILED = 'failed';
+    public const STATUS_SUCCEEDED = 'succeeded';
```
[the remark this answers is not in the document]
egrojMonroy: Moved to delayed job and updates all references

app/Jobs/InsertGeojsonToDBJob.php
```diff
@@ -15,94 +15,76 @@
...
 
         } catch (Exception $e) {
             Log::error('Error in InsertGeojsonToDBJob: ' . $e->getMessage());
-            DelayedJob::where('uuid', $this->job_uuid)->update([
+            DelayedJob::where('id', $this->delayed_job_id)->update([
                 'status' => self::STATUS_FAILED,
                 'payload' => ['error' => $e->getMessage()],
                 'status_code' => Response::HTTP_INTERNAL_SERVER_ERROR,
             ]);
+            Redis::del($this->redis_key);
```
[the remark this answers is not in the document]
egrojMonroy: The best way, moved already.

app/Jobs/InsertGeojsonToDBJob.php
```diff
@@ -15,94 +15,76 @@
...
-                'uuid' => $this->job_uuid,
-                'status' => self::STATUS_PENDING,
-                'created_at' => now(),
-            ]);
-            $uuids = $service->insertGeojsonToDB(
-                $this->geojsonFilename,
+            $delayedJob = DelayedJob::findOrFail($this->delayed_job_id);
+            if(!$delayedJob) {
+                throw new \Exception('DelayedJob not found', Response::HTTP_NOT_FOUND);
+            }
```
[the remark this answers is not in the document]
egrojMonroy: Removed the delayedJobs.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
egrojMonroy: restructure the code without creating the job.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
egrojMonroy: done

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -583,29 +588,34 @@ public function uploadGeoJSONFile(Request $request)
...
-            $geojsonContent = file_get_contents($filePath);
-            $polygonLoaded = $this->GetAllPolygonsLoaded($geojsonContent, $site_id);
+            $polygonLoaded = $this->GetAllPolygonsLoaded($geojson_content, $site_id);
 
             return response()->json($polygonLoaded->original, 200);
         }
 
         if ($submitPolygonsLoaded) {
-            $job = new InsertGeojsonToDBJob($filename, $site_id, 'site', $body['primary_uuid'] ?? null, $body['submit_polygon_loaded']);
+          $job = new InsertGeojsonToDBJob($redis_key, $delayedJob->id,  $site_id, 'site', $body['primary_uuid'] ?? null, $body['submit_polygon_loaded']);
```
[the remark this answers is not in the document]
egrojMonroy: restructured

app/Helpers/CreateVersionPolygonGeometryHelper.php
```diff
@@ -53,7 +48,7 @@ public static function createVersionPolygonGeometry(string $uuid, $geometry)
 
             return response()->json(['message' => 'Failed to create site polygon version.'], 500);
         } catch (\Exception $e) {
-            return response()->json(['error' => 'An error occurred: ' . $e->getMessage()], 500);
+            throw $e;
```
roguenet, line 51: Was this for testing? If you're intentionally letting the exception bubble up to the HTTP layers, then there's no reason for the try / catch.

app/Http/Controllers/V2/Terrafund/TerrafundClipGeometryController.php
```diff
@@ -23,10 +24,11 @@ public function clipOverlappingPolygonsBySite(string $uuid)
         ini_set('memory_limit', '-1');
         $user = Auth::user();
         $polygonUuids = GeometryHelper::getSitePolygonsUuids($uuid)->toArray();
-        $job = new FixPolygonOverlapJob($polygonUuids, $user->id);
+        $delayedJob = DelayedJob::create(['status' => DelayedJob::STATUS_PENDING]);
```
roguenet, line 27: It's a little weird to me that setting `status` is required. Can we make it a default value on the DB column definition instead? I don't think we'd ever have a use case where we create a delayed job in another status.

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
```diff
@@ -223,30 +223,38 @@ function ($attribute, $value, $fail) {
...
+            return response()->json($polygonLoaded->original, 200);
         }
 
-        if ($submitPolygonsLoaded) {
-          $job = new InsertGeojsonToDBJob($redis_key, $delayedJob->id, $site_id, 'site', $body['primary_uuid'] ?? null, $body['submit_polygon_loaded']);
+        $redis_key = 'kml_file_' . uniqid();
+        Redis::set($redis_key, $geojsonContent, 'EX', 7200);
+        $delayedJob = DelayedJob::create(['status' => DelayedJob::STATUS_PENDING]);
+    
+        if($delayedJob) {
```
roguenet, line 240: I think this check is superfluous. There isn't a reasonable expectation that creating a row in this DB table could fail (and please fix in the lower instances of this pattern as well)

app/Http/Controllers/V2/Terrafund/TerrafundClipGeometryController.php
egrojMonroy: Made pending as default. changed the creation for all ocurrences

app/Http/Controllers/V2/Terrafund/TerrafundCreateGeometryController.php
egrojMonroy: removed

app/Helpers/CreateVersionPolygonGeometryHelper.php
egrojMonroy: try catch removed

roguenet (approved):

roguenet (approved):

PR #533, opened 2024-10-31
[TM-1303] Socioeconomic Partners

roguenet:
https://gfw.atlassian.net/browse/TM-1303
[FE PR](https://github.com/wri/wri-terramatch-website/pull/608)


app/Console/Commands/OneOff/FixDuplicateDemographics.php
```diff
@@ -29,32 +29,34 @@ public function handle()
...
+                    $workdays[$demographic->demographical_id][] = $demographic->toArray();
                 }
 
                 return true;
             });
 
         foreach ($workdays as $workdayId => $stats) {
             foreach ($stats as $stat) {
-                $workday = Workday::find($workdayId);
+                $workday = Workday::withTrashed()->find($workdayId);
```
Scriptmatico, line 59: Why is withTrashed neccessary here?

roguenet: Because the query that pulls demographics above can pull demographics in trashed workdays, and since the code is already here to fix them, it seems worthwhile to fix them in case the workdays are ever resurrected.

Scriptmatico (approved):

PR #534, opened 2024-10-31
[TM-1386] update reports statuses

egrojMonroy:
This is a command that will update via projects projectReport, siteReport and tasks from 
“awaiting approval” or “needs more information” to “approved”.


cesarLima1 (approved):

roguenet (changes requested):
Two thing:
* This would be better implemented as a Command than a seeder, as it more conceptually aligns with what this tool is doing. Seeders are for getting initial data into a table, not for modifications
* I don't think the .csv makes much sense to check in. Would be better to just upload to the EC2 box and pass it as an argument to the command. 

`BulkWorkdayImport` is an example of a Command that works in this way.

roguenet (commented):
the command looks good, but please remove the seeder and checked in .csv.

roguenet (approved):

PR #536, opened 2024-10-31
[TM-1432] dashboard jobs created controller moved to jobs

cesarLima1:
https://gfw.atlassian.net/browse/TM-1432


roguenet:
If we're going to treat these improvements as belonging to an epic, let's spin up a real epic branch. If we don't need to do it that way, let's merge directly into `staging`. This `test_cache` branch should not be part of the real process for getting this code live.

pachonjcl (approved):

PR #556, opened 2024-11-07
[TM-1343] Change filters for map

egrojMonroy:


app/Helpers/TerrafundDashboardQueryHelper.php
```diff
@@ -135,8 +135,8 @@ public static function getPolygonsByStatusOfProjects($request)
         $projectUuids = TerrafundDashboardQueryHelper::buildQueryFromRequest($request)
             ->pluck('v2_projects.uuid');
 
-        $requestedStatuses = $request->input('statuses');
+        $approvedStatus = ['approved'];
```
pachonjcl, line 138: why are you using an array here?

app/Http/Controllers/V2/Dashboard/ViewProjectController.php
```diff
@@ -19,11 +19,8 @@ public function getIfUserIsAllowedToProject(String $uuid)
         /** @var User $user */
         $user = Auth::user();
         if ($user->hasRole('government')) {
-            $isAllowed = Project::where('uuid', $uuid)
-                ->where('country', $user->country)
-                ->exists();
             $response = (object)[
-                'allowed' => $isAllowed,
+              'allowed' => false,
```
pachonjcl, line 23: is the endpoint still needed now?

pachonjcl (changes requested):

cesarLima1 (approved):

app/Helpers/TerrafundDashboardQueryHelper.php
egrojMonroy: Because the other function to retrieve data, receives an array of all statuses needed. 
Whay should I do in that case?

app/Http/Controllers/V2/Dashboard/ViewProjectController.php
egrojMonroy: yes, but for government now the request is that it cannot see any polygons, that is why is always false. 
Or should I check from FE ?

pachonjcl (approved):

PR #566, opened 2024-11-13
[TM-1352] add logic to get new fields results

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-1352)

### DIFF
--- a/app/Models/V2/Projects/ProjectReport.php
+++ b/app/Models/V2/Projects/ProjectReport.php
@@ -217,6 +217,15 @@ class ProjectReport extends Model implements MediaModel, AuditableContract, Repo
             Workday::COLLECTION_PROJECT_DIRECT,
             Workday::COLLECTION_PROJECT_CONVERGENCE,
         ],
+        'direct' => [
+            Workday::COLLECTION_PROJECT_DIRECT,
+        ],
+        'convergence' => [
+            Workday::COLLECTION_PROJECT_CONVERGENCE,
+        ],
+        'non-tree' => [
+            Workday::COLLECTION_PROJECT_NON_TREE,
+        ],
     ];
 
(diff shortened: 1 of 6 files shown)


roguenet:
When this comes out of draft, please request review from me.

app/Models/V2/Projects/ProjectReport.php
```diff
@@ -439,11 +445,43 @@ public function getWorkdaysTotalAttribute(): int
...
 
+    public function getWorkdaysDirectTotalAttribute(): int
+    {
+        return $this->workdays_direct ?? 0;
+    }
+
+    public function getWorkdaysConvergenceTotalAttribute(): int
+    {
+        return $this->workdays_convergence ?? 0;
+    }
```
roguenet, line 456: I think these two are superfluous. The resource can just use `workdays_direct` and `_convergence` directly, which would be in line with how the other totals are accessed. Also, since these are using SQL `sum()`, I believe that they'll always return 0 automatically if there are no matching records.

app/Models/V2/Projects/ProjectReport.php
```diff
@@ -439,11 +445,43 @@ public function getWorkdaysTotalAttribute(): int
...
+
+    public function getNonTreeTotalAttribute(): int
+    {
+        if (empty($this->task_id)) {
+            return 0;
+        }
+
+        return TreeSpecies::where('speciesable_type', SiteReport::class)
+            ->whereIn('speciesable_id', $this->task->siteReports()->select('id'))
```
roguenet, line 465: Should this be including all site reports? Typically these types of queries are report status dependent. For instance `getWorkdaysTotalAttribute` in this class uses 
```
->whereIn('workdayable_id', $this->task->siteReports()->hasBeenSubmitted()->select('id'))
```

PR #585, opened 2024-11-22
[TM-1474] approved filter to high level metrics in all entities

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-1474)


tests/Unit/Models/V2/Projects/ProjectTest.php
```diff
@@ -148,7 +148,9 @@ public function test_workday_count()
         Demographic::factory()->create(['demographical_id' => $workday->id, 'amount' => 19]);
 
         // 42 = 5 and 7 from the approved site's reports and 13 and 17 from the project reports
```
pachonjcl, line 150: This comment should either be updated or deleted, 42 is no longer valid.

pachonjcl (changes requested):

pachonjcl (approved):

PR #612, opened 2024-12-13
[TM-1467] update requirements file

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-1467)


resources/python/polygon-indicator/requirements.txt
```diff
@@ -1,11 +1,11 @@
...
 fiona==1.10.1
 exactextract==0.2.0
 rasterio==1.4.3
-boto3==1.35.77
-gdal==3.6.2
-pyyaml==6.0.1
+gdal==3.4.1
+pyyaml==5.4.1
 rasterstats==0.20.0
+boto3==1.28.17
```
roguenet, line 11: Almost all of these updates are downgrades. Is that intentional?

resources/python/polygon-indicator/requirements.txt
```diff
@@ -1,11 +1,11 @@
-shapely==2.0.1
-geopandas==1.9.0
-pandas==2.1.2
+shapely==2.0.6
+geopandas==1.0.1
```
roguenet, line 2: This one in particular worries me - that seems like it's a pretty big downgrade.

LimberHope: I understand your concern, but the version 1.0.1 of geopandas is the most recent stable release. The 1.9.0 version you mentioned was actually a typo, which might have caused some confusion. The version 1.0.1 is the correct one to use.

roguenet (approved):

PR #632, opened 2024-12-20
[TM-1531] remove entity from delayedjobs and create metadata

egrojMonroy:


app/Models/DelayedJob.php
```diff
@@ -18,17 +18,12 @@ class DelayedJob extends Model
 
     protected $table = 'delayed_jobs';
 
-    protected $fillable = ['uuid', 'status', 'status_code', 'payload', 'entity_type', 'entity_id', 'created_by', 'is_acknowledged', 'name'];
+    protected $fillable = ['uuid', 'status', 'status_code', 'payload', 'metadata', 'created_by', 'is_acknowledged', 'name'];
 
     protected $casts = [
         'uuid' => 'string',
```
roguenet, line 24: If you add `'metadata' => 'json'` here, Laravel will handle the serialization for you, and you won't need to `json_encode` or `json_decode` when accessing metadata.

egrojMonroy: changed to array, as that worked.

roguenet (approved):

PR #677, opened 2025-02-01
[TM-1681] Merge workdays and restoration partners

roguenet:
https://gfw.atlassian.net/browse/TM-1681

This ticket's main purpose is to merge merge the workdays and restoration partner tables into a single demographics table with a differentiating `type` column. We'll be adding more demographic types in the near future, and this prepares for that work.

Note that I updated `mysql-schema.dump` and removed a ton of old migrations. It was time to do this anyway, but in this particular case it was necessary because the migrations that were running in the test suite were running into problems with multiple instances of renaming `demographics` tables over time. The schema dump matches what's currently on production. I left all migrations from the beginning of December onward in case there were any devs that hadn't run them recently on their local machines.


pachonjcl (approved):
Lot of removed files, I was afraid about the number of files.

LimberHope (approved):

PR #717, opened 2025-02-28
[TM-1759] add polygons manage to admins

egrojMonroy:


roguenet:
I think this isn't appropriate because it gives admins of one framework access to the polygons in all frameworks (unless the policy in v3 is going to filter on framework, and if it is, why is a new permission needed in the first place?)

roguenet (commented):
This has the same problem: it gives all admins access to all polygons. I don't see why we need to change the policies in this codebase at all. Please see my comments on the v3 policy PR.

PR #721, opened 2025-03-01
[TM-1766] entity get and index site

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-1766)


routes/api_v2.php
```diff
@@ -330,7 +329,6 @@
     });
 
     Route::prefix('sites')->group(function () {
-        Route::get('/', AdminIndexSitesController::class);
```
roguenet, line 333: This should also be removing the `{project}/sites` route, and comment on `ViewEntityController` should be updated to note that Site is also no longer being fetched from v2. (and associated documentation for those routes should be removed)

roguenet (commented):
Looks like the `ViewProjectSitesController` (which this PR should be removing) has some additional sorting / filtering requirements that will need to be supported in v3. Please check that they're all actually needed / used before putting in the effort to implement them.

roguenet:
Looks like there are some specs that need to be removed.

roguenet (approved):

roguenet (approved):

PR #728, opened 2025-03-06
[TM-1718] fix query

egrojMonroy:


roguenet (approved):

app/Services/Dashboard/JobsCreatedService.php
```diff
@@ -14,15 +14,19 @@ public function calculateJobsCreated(Request $request)
     {
         $query = TerrafundDashboardQueryHelper::buildQueryFromRequest($request);
 
-        $projectIds = $query->select('v2_projects.id', 'organisations.type');
+        $projectIds = $query->select('v2_projects.id');
+        $projectReportIds = ProjectReport::whereIn('project_id', $projectIds)
+        ->where('status', 'approved')
+        ->pluck('id');
```
roguenet, line 20: This should still use `select('id')` instead of `pluck('id')`. That will use a subquery instead of loading all the project report ids into memory just to turn around and send them back to the DB.

roguenet (changes requested):

app/Services/Dashboard/JobsCreatedService.php
egrojMonroy: changed done

roguenet (approved):

PR #755, opened 2025-03-20
[TM-1768] remove project reports files and definitions

LimberHope:
Task Link


app/Http/Controllers/V2/Tasks/ViewTaskUuidController.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace App\Http\Controllers\V2\Tasks;
+
+use App\Http\Controllers\Controller;
+use App\Models\V2\Tasks\Task;
+
+class ViewTaskUuidController extends Controller
```
roguenet, line 8: I didn't even realize when I was reading over the FE PR that this is a whole new endpoint. Definitely should be replaced with work in the v3 repo in that case.

roguenet (approved):

roguenet (approved):

roguenet (approved):

PR #776, opened 2025-03-26
[TM-1898] Improve docker

roguenet:
https://gfw.atlassian.net/browse/TM-1898

I went down this rabbit hole because the current docker build wasn't working locally, and I had to blow away and rebuild my local docker containers today. Highlights:
* Update to using `docker compose` over the old (now unsupported) `docker-compose`
* Update to using `docker buildx bake` for the initial build of the compose file. 
* I had to bump the version of GDAL being created in the docker container to 3.5 to deal with some python dependency version issues. This _does not_ affect the EC2 boxes that are running our services on Amazon, and the unit tests pass, so I think this is OK for now.
* Use caching in github actions to try to avoid the full build on pull requests
  * After the last commit, this is finally working. Note how the last run on this PR only took ~6 minutes instead of 25+ minutes.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -5,15 +5,10 @@ jobs:
   lintTest:
     runs-on: ubuntu-latest
     steps:
-      - name: Build and push
-        uses: docker/build-push-action@v6
-        with:
-          context: .
-          push: false
-          cache-from: type=gha
-          cache-to: type=gha,mode=max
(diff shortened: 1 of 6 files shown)


roguenet:
I finally have the docker build cache working. I'll look into the spec failures tomorrow - these same specs are failing for me locally, and I don't understand why they used to pass when the container was being built fully fresh each time.

cesarLima1 (approved):

PR #780, opened 2025-03-28
[TM-1869] better handling of failed job

egrojMonroy:
This is a handling for failed jobs and update the code in delayedJobs table on any failure.


app/Providers/AppServiceProvider.php
```diff
@@ -36,5 +41,41 @@ public function boot()
 
         SitePolygon::observe(SitePolygonObserver::class);
         Media::observe(MediaObserver::class);
+        Queue::failing(function (JobFailed $event) {
+            $jobName = $event->job->resolveName();
+            $jobId = $event->job->uuid(); // Get UUID if available
+            $exceptionMessage = $event->exception->getMessage();
+
+            $rawBody = $event->job->getRawBody();
+            ;
```
roguenet, line 50: 🛅

app/Providers/AppServiceProvider.php
```diff
@@ -36,5 +41,41 @@ public function boot()
...
+                    $delayedJobId = null;
+                }
+
+            } catch (\Exception $e) {
+                Log::error('Error deserializing command: ' . $e->getMessage());
+
+                return;
+            }
+
+            $delayedJob = DelayedJob::where('id', $delayedJobId)->first();
```
roguenet, line 71: Since `delayedJobId` can be null, it's likely worth filtering on that
```
$delayedJob = $delayedJobId == null ? null : DelayedJob::where(...)
```

app/Providers/AppServiceProvider.php
egrojMonroy: removed

roguenet (approved):

PR #816, opened 2025-04-16
[TM-1933] Avoid send emails when site report is submitted

Scriptmatico:
https://gfw.atlassian.net/browse/TM-1933

### DIFF
--- a/app/Http/Controllers/V2/Entities/SubmitEntityWithFormController.php
+++ b/app/Http/Controllers/V2/Entities/SubmitEntityWithFormController.php
@@ -12,6 +12,7 @@ use App\Models\V2\UpdateRequests\UpdateRequest;
 use App\StateMachines\ReportStatusStateMachine;
 use Illuminate\Http\JsonResponse;
 use Illuminate\Http\Request;
+use App\Models\V2\Sites\SiteReport;
 
 class SubmitEntityWithFormController extends Controller
 {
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/V2/Entities/SubmitEntityWithFormController.php
```diff
@@ -41,7 +42,10 @@ public function __invoke(EntityModel $entity, Request $request)
             $this->saveAuditStatusProjectDeveloperSubmitNotUpdateRequest($entity);
         }
 
-        SendProjectManagerJobs::dispatch($entity);
+        if (get_class($entity) !== SiteReport::class) {
+            SendProjectManagerJobs::dispatch($entity);
+        }
```
roguenet, line 47: Please check with Claire and see if this should also apply to project reports and nursery reports.

Scriptmatico: Sure let me ask her and I will be back here

roguenet:
Looks like the linter found a failure. You should be able to fix automatically with `make lint-fix`

app/Http/Controllers/V2/Entities/SubmitEntityWithFormController.php
Scriptmatico: Claire confirmed that is needed for nursery report too, can you please review again?

roguenet (approved):

PR #826, opened 2025-04-23
[TM-1962] Clean conditional data admin edit

roguenet:
https://gfw.atlassian.net/browse/TM-1962

There were a couple of problems here:
* The default value for a form field that has been hidden is not some "empty" value (like `0`, `""` or `false`), it's null. Especially in the case of numerical fields, this has real consequences, as 0 can mean something quite different from null.
* There was a bug in the calculation for which form fields to clear out; it was using `$question` from the block above to pull the linked field key info, instead of `$child`.
* When an admin edits an already approved form, we want to apply the data cleaning right then, as there won't be the normal opportunity to do it later as part of the approval process.

I also updated the one off that runs the cleaning process on all approved reports to give better feedback and progress reporting. It will need to be run again with the next release.

### DIFF
--- a/app/Models/Traits/UsesLinkedFields.php
+++ b/app/Models/Traits/UsesLinkedFields.php
@@ -98,6 +98,13 @@ trait UsesLinkedFields
         }
 
         $this->update($entityProps);
+
+        if ($this->status == EntityStatusStateMachine::APPROVED) {
+            // This state only occurs when an admin is editing an already approved entity. Under these circumstances,
+            // we'd rather see the entity saved with the clean data view than preserve any temporarily set data the
+            // admin might have put in place.
+            $this->cleanConditionalAnswers($form);
+        }
     }
 
     public function calculateCompletion(Form $form): int
--- a/app/Models/Traits/UsesLinkedFields.php
+++ b/app/Models/Traits/UsesLinkedFields.php
@@ -5,6 +5,7 @@ namespace App\Models\Traits;
 use App\Models\Interfaces\HandlesLinkedFieldSync;
 use App\Models\V2\Forms\Form;
 use App\Models\V2\Forms\FormQuestion;
+use App\StateMachines\EntityStatusStateMachine;
 
 trait UsesLinkedFields
 {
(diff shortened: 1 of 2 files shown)


app/Models/Traits/UsesLinkedFields.php
```diff
@@ -339,13 +347,14 @@ public function cleanConditionalAnswers(Form $form): void
 
         foreach ($childQuestions as $child) {
             if (in_array($child->parent_id, $conditionalsToBeCleaned)) {
-                $fieldConfig = data_get($fieldsConfig, $question->linked_field_key);
+                $fieldConfig = data_get($fieldsConfig, $child->linked_field_key);
                 $property = data_get($fieldConfig, 'property', null);
                 if ($this->isPlainField($child->input_type)) {
-                    $entityProps[$property] = $this->getDefaultValue($child->input_type);
+                    $entityProps[$property] = null;
```
roguenet, line 353: The default value for a hidden child question should always be null, not a faked "empty" value. This is especially important for number fields where seeing a null vs a 0 in the exported data can have very different connotations.

app/Models/Traits/UsesLinkedFields.php
```diff
@@ -339,13 +347,14 @@ public function cleanConditionalAnswers(Form $form): void
 
         foreach ($childQuestions as $child) {
             if (in_array($child->parent_id, $conditionalsToBeCleaned)) {
-                $fieldConfig = data_get($fieldsConfig, $question->linked_field_key);
+                $fieldConfig = data_get($fieldsConfig, $child->linked_field_key);
```
roguenet, line 350: This bug meant this code wasn't really doing much of anything useful. I'm surprised it passed product review.

Scriptmatico: Is interesting how this was not detected, maybe a linter or something in the IDE can prevent this?

app/Models/Traits/UsesLinkedFields.php
Scriptmatico, line 353: I'm just wondering in this line if there is on a potential issue, If $property is null, you still do:

`$entityProps[$property] = $value;`

which results in $entityProps[null] = $value, this is dangerous because it inserts an undefined key into the array.

Scriptmatico (approved):
Despite of my comment this looks good to me.

LimberHope (dismissed):

app/Models/Traits/UsesLinkedFields.php
roguenet: `$question` is technically a valid variable in the current scope, from the block above. The problem here was that it was running `data_get` on the same linked field key for every child, instead of using the child's key. I don't think this is anything a linter could catch.

PR #834, opened 2025-04-30
[TM-1958] GADM codes for countries / states / districts

roguenet:
https://gfw.atlassian.net/browse/TM-1958


app/Console/Commands/OneOff/MigrateLocationCodes.php
```diff
@@ -0,0 +1,239 @@
...
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Moves historical state / country codes in the DB to the official GADM codes for those places.';
+
+    protected const DATA_API = 'https://data-api.globalforestwatch.org/dataset/gadm_administrative_boundaries/v4.1.85/query';
+    protected const API_KEY = '349c85f6-a915-4d3d-b337-0a0dafc0e5db';
```
Scriptmatico, line 34: Should not this go in to environment variables?

app/Console/Commands/OneOff/MigrateLocationCodes.php
```diff
@@ -0,0 +1,239 @@
...
+        $this->info("\n\nUpdating leadership nationalities...");
+        foreach (Leaderships::whereNotNull('nationality')->get() as $leadership) {
+            $leadership->update(['nationality' => $this->findCountry($leadership->nationality)]);
+        }
+    }
+
+    protected function handleMigrationSet($query, $definitions)
+    {
+        $count = (clone $query)->count();
+        $this->withProgressBar($count, function ($progressBar) use ($query, $definitions, &$chunks) {
```
Scriptmatico, line 121: I'm not sure why &$chunks is declared but never used.

app/Console/Commands/OneOff/MigrateLocationCodes.php
```diff
@@ -0,0 +1,239 @@
...
+                                }
+                            }
+
+                            if (Str::endsWith($type, '_single')) {
+                                $entity[$column] = data_get($values, 0);
+                            } else {
+                                $entity[$column] = $values;
+                            }
+                        }
+                    }
```
Scriptmatico, line 150: There is too many if conditions in this block, I can see some small improvements, you can store the _single and gadm_0 in to variables to use to validate below. I think this snippet improves the readability and maintainability.

```suggestion
                    $isSingle = Str::endsWith($type, '_single');
                    $isCountry = Str::startsWith($type, 'gadm_0');

                    foreach ($columns as $column) {
                        $values = $isSingle ? [$entity[$column]] : $entity[$column];
                        
                        $values = collect($values)
                            ->map(fn ($v) => $isCountry ? $this->findCountry($v) : $this->findState($v))
                            ->filter()
                            ->toArray();

                        $entity[$column] = $isSingle ? data_get($values, 0) : $values;
                    }
```

app/Console/Commands/OneOff/MigrateLocationCodes.php
roguenet: Not for this one-off script; I didn't see a need to do that. It's not secret, and we're not doing a deeper integration in PHP at this time.

app/Console/Commands/OneOff/MigrateLocationCodes.php
roguenet: I'm going to apply this suggestion. However, just a note for the future: I typically don't get fussy over the formatting, readability or maintainability of one-off scripts. They don't need to be maintainable, they only get run once. In fact, I should pull together a tech debt ticket soon to clear out some old ones. In this particular case, I tested the heck out of this code myself, and with a refactor, while it looks clean to me, I'm going to have to do some more testing to make sure it still works correctly.

app/Console/Commands/OneOff/MigrateLocationCodes.php
roguenet: Good catch. Was the result of a refactor. I"m surprised the linter didn't complain about it. Fixed now.

app/Console/Commands/OneOff/MigrateLocationCodes.php
roguenet: I'm actually going to need to roll this back. The formatting wasn't right, although I don't understand why from this diff, and the the script didn't work correctly anymore. I don't think it's worth spending the time to try to figure out why when the previous version has already been thoroughly tested.

cesarLima1 (approved):

Scriptmatico (approved):

PR #971, opened 2025-08-04
[TM-2343] add command to disable fundo flora funding programmes

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-2343)

### DIFF
--- a/app/Console/Commands/CloseFundoFloraFundingProgrammesCommand.php
+++ b/app/Console/Commands/CloseFundoFloraFundingProgrammesCommand.php
@@ -1,0 +1,104 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Models\V2\FundingProgramme;
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\Log;
+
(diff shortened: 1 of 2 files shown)


app/Console/Commands/CloseFundoFloraFundingProgrammesCommand.php
```diff
@@ -0,0 +1,104 @@
...
+     * Execute the console command.
+     */
+    public function handle()
+    {
+        $this->info('Starting to close Fundo Flora funding programmes...');
+
+        // Fundo Flora funding programme UUIDs
+        $fundoFloraUuids = [
+            'e1bdd11c-aefe-4424-ad57-0e8c5ea1e2ab', // Non-profit funding programme
+            '22cdd788-abab-4542-ab6f-d8999aacab9d', // Enterprise funding programme
```
roguenet, line 35: Did you get these programme UUIDs from staging or prod? They might not be the same (although we've updated prod to staging recently enough that they _probably_ are).

PR #1020, opened 2025-09-04
[TM-2460] add validation to update organisation

LimberHope:
[Task link](https://gfw.atlassian.net/browse/TM-2460)

### DIFF
--- a/app/Models/V2/FinancialIndicators.php
+++ b/app/Models/V2/FinancialIndicators.php
@@ -137,10 +137,12 @@ class FinancialIndicators extends Model implements MediaModel, HandlesLinkedFiel
         if (! empty($organisationId) && empty($financialReportId)) {
             if ($startMonth !== null || $currency !== null) {
                 $organisation = Organisation::isUuid($organisationId)->first();
-                $organisation->update([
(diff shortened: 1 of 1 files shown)


app/Models/V2/FinancialIndicators.php
```diff
@@ -137,10 +137,12 @@ public static function syncRelation(Model $entity, string $property, string $inp
...
-                $organisation->update([
-                    'fin_start_month' => $startMonth,
-                    'currency' => $currency,
-                ]);
+                if ($organisation) {
+                    $organisation->update([
+                        'fin_start_month' => $startMonth,
+                        'currency' => $currency,
+                    ]);
+                }
```
roguenet, line 145: Once we have the bug fully isolated, I'd also like a small logic fix for this block and the one above. Should look something more like:
```
if ($startMonth !== null || $currency !== null) {
  if (! empty($financialReport)) {
    // update FR
  } else if (! empty($organisationId)) {
    // find and update org
  }
}
```

LimberHope: resolved

PR #1045, opened 2025-09-16
[TM-2408] Add command to update planting status from CSV for project and site reports

Scriptmatico:


roguenet (commented):
Please don't commit the CSV to source control. The standard practice for these one-offs is to write the script to accept the CSV from the command line and upload the file to the EC2 with SSH. See `ImportTreeSpeciesAssociations` for an example.

PR #1048, opened 2025-09-17
[TM-2485] Add two new project pitch fields for Fundo Flora

roguenet:
https://gfw.atlassian.net/browse/TM-2485


roguenet:
Nope, simply including them in the linked fields is all that's needed make it possible to include them in forms.

